PR #13, opened 2023-01-10
Improve perfs on noisy fetch chains

ThibaudDauce:
Since fetchChains is often an IO intensive task, call the futures in parallel instead of calling them one by one can save a lot of time.

This implementation still call in sync for each keyword, we may put the outer loop inside the `join_all(futures)` too to call everything in parallel

We may have problems about the number of requests in parallel in the browser… Currently we send around 900 fetchChains requests in parallel (if it's HTTP, the browser could be not happy at all…)


src/core/search.rs
```diff
@@ -238,14 +239,24 @@ pub trait FindexSearch<
...
-                    .fetch_chain_table(&HashSet::from([uid.clone()]))
-                    .await?;
+
+            // Fetch all chain table values one by one to increase noise.
+            let chain_table_uids_hashset: Vec<_> = chain_table_uids
+                .iter()
+                .map(|uid| HashSet::from([uid.clone()]))
+                .collect();
+
+            let mut futures = vec![];
```
tbrezot, line 249: Can you initialize this with a capacity ? If there is not too many misses, the actual length should be close to `chain_table_uids_hashset.len()`.

src/error.rs
```diff
@@ -17,7 +17,6 @@ pub enum FindexErr {
     Other(String),
 
     // Findex implementation for FFI
-    #[cfg(feature = "interfaces")]
```
tbrezot, line 20: Why do you remove these lines ? It does not seem to be related to this PR's topic.

src/core/search.rs
ThibaudDauce: Done

src/error.rs
ThibaudDauce: I don't really know why my changes requires this but the compiler is unhappy :-D `FindexErr::Callback` is used inside `noisy_fetch_chains` which may be used outside the `interface` feature? I don't know… Maybe something with the removal of `futures` as an optional dependency?

ThibaudDauce: It was a problem on the `improve_errors` base branch, fixed!

src/interfaces/ffi/api.rs
```diff
@@ -181,6 +185,15 @@ pub unsafe extern "C" fn h_search(
         )
     };
 
+    let fetch_chains_batch_size = if fetch_chains_batch_size <= 0 {
+        SECURE_FETCH_CHAINS_BATCH_SIZE
+    } else {
+        ffi_unwrap!(
+            usize::try_from(fetch_chains_batch_size),
+            "loop_iteration_limit must be a positive int"
```
Manuthor, line 193:
```suggestion
            "fetch_chains_batch_size must be a positive int"
```

src/core/mod.rs
```diff
@@ -14,6 +14,8 @@
 const ENTRY_TABLE_KEY_DERIVATION_INFO: &[u8] = b"Entry Table key derivation info.";
 const CHAIN_TABLE_KEY_DERIVATION_INFO: &[u8] = b"Chain Table key derivation info.";
 
+pub const SECURE_FETCH_CHAINS_BATCH_SIZE: usize = 1;
```
tbrezot, line 17: Move this to `src/interfaces/generic_parameters.rs`

src/core/search.rs
```diff
@@ -238,14 +252,24 @@ pub trait FindexSearch<
...
             let kwi_value: DemScheme::Key = kwi.derive_dem_key(CHAIN_TABLE_KEY_DERIVATION_INFO);
             let mut chain = Vec::with_capacity(chain_table_uids.len());
-            for uid in chain_table_uids {
-                // Fetch all chain table values one by one to increase noise.
-                let encrypted_item = self
-                    .fetch_chain_table(&HashSet::from([uid.clone()]))
-                    .await?;
+
+            // Fetch all chain table values one by one to increase noise.
+            let chain_table_uids_hashset: Vec<HashSet<_>> = chain_table_uids
```
tbrezot, line 257:
```suggestion
            let chain_table_uids_hashset: Vec<_> = chain_table_uids
```

src/core/search.rs
```diff
@@ -238,14 +252,24 @@ pub trait FindexSearch<
         for (kwi, chain_table_uids) in kwi_chain_table_uids.iter() {
             let kwi_value: DemScheme::Key = kwi.derive_dem_key(CHAIN_TABLE_KEY_DERIVATION_INFO);
             let mut chain = Vec::with_capacity(chain_table_uids.len());
-            for uid in chain_table_uids {
-                // Fetch all chain table values one by one to increase noise.
-                let encrypted_item = self
-                    .fetch_chain_table(&HashSet::from([uid.clone()]))
-                    .await?;
+
+            // Fetch all chain table values one by one to increase noise.
```
tbrezot, line 256:
```suggestion
            // Fetch all chain table values by batch of `batch_size` to increase noise.
```

src/interfaces/ffi/api.rs
```diff
@@ -147,6 +150,7 @@ pub unsafe extern "C" fn h_search(
     keywords_ptr: *const c_char,
     max_results_per_keyword: c_int,
     max_depth: c_int,
+    fetch_chains_batch_size: c_int,
```
tbrezot, line 153: can we use a `c_uint` here ? and use `0` to trigger the default in the test below.

src/interfaces/ffi/api.rs
```diff
@@ -181,6 +185,15 @@ pub unsafe extern "C" fn h_search(
         )
     };
 
+    let fetch_chains_batch_size = if fetch_chains_batch_size <= 0 {
+        SECURE_FETCH_CHAINS_BATCH_SIZE
+    } else {
+        ffi_unwrap!(
+            usize::try_from(fetch_chains_batch_size),
+            "fetch_chains_batch_size must be a positive int"
```
tbrezot, line 193: I am not sure this error message is adequate since we treated all negative cases above. Use back-quotes around the parameter name.

src/interfaces/wasm_bindgen/api.rs
```diff
@@ -36,8 +38,9 @@ pub async fn webassembly_search(
     master_key: Uint8Array,
     label_bytes: Uint8Array,
     keywords: ArrayOfKeywords,
-    max_results_per_keyword: usize,
-    max_depth: usize,
+    max_results_per_keyword: i32,
+    max_depth: i32,
```
tbrezot, line 42: Why did you change these types?

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Changed
+
+- Add `fetch_chains_batch_size` argument to search
+- `fetch_chains` calls are now run in parallel in wasm (not working in Java/Flutter for now)
```
tbrezot, line 10:
```suggestion
- in `search`, `fetch_chains` calls are now run in parallel for batches of `fetch_chains_batch_size` (not working for the Java/Flutter interfaces for now)
```

src/interfaces/wasm_bindgen/api.rs
ThibaudDauce: Because unsigned types doesn't have a really good representation in JS so I prefer the interface to match the real types of JS to not be surprise, for example, with the previous types, I didn't know what would happen if the JS sent a `-1` (since the TS type is just `number`)…

src/interfaces/ffi/api.rs
ThibaudDauce: I changed it but same as in JS… For example Flutter doesn't have unsigned integer so the user can send a `-1`

Manuthor (approved):

PR #17, opened 2023-02-02
Findex cloud

ThibaudDauce:


src/interfaces/ffi/api.rs
```diff
@@ -387,7 +289,9 @@ pub unsafe extern "C" fn h_compact(
         fetch_all_entry_table_uids: Some(fetch_all_entry_table_uids),
     };
 
-    ffi_unwrap!(executor::block_on(findex.compact(
+    let rt = ffi_unwrap!(tokio::runtime::Runtime::new());
```
Manuthor, line 292: for my knowledge only, was there something wrong with `executor::block_on`? now we need tokio for findex.

src/interfaces/ffi/api.rs
```diff
@@ -396,3 +300,295 @@ pub unsafe extern "C" fn h_compact(
...
+/// - `indexed_values`          : (output) search result
+/// - `master_key`              : master key
+/// - `label`                   : additional information used to derive Entry
+///   Table UIDs
+/// - `keywords`                : `serde` serialized list of base64 keywords
+/// - `max_results_per_keyword` : maximum number of results returned per keyword
+/// - `max_depth`               : maximum recursion depth allowed
+/// - `progress_callback`       : callback used to retrieve intermediate results
+///   and transmit user interrupt
+/// - `fetch_entry`             : callback used to fetch the Entry Table
```
Manuthor, line 327: parameters-description does not match

src/interfaces/wasm_bindgen/api.rs
```diff
@@ -123,3 +125,147 @@ pub async fn webassembly_upsert(
         .await
         .map_err(|e| JsValue::from(format!("During Findex upsert: {e}")))
 }
+
+/// See [`FindexSearch::search()`](crate::core::FindexSearch::search).
+///
+/// # Parameters
+///
+/// - `master_key`              : master key
```
Manuthor, line 133: parameters-description does not match

Manuthor (commented):
some details, otherwise all good

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+                )));
+            }
+        }
+
+        Ok(token)
+    }
+}
+
+impl Token {
+    pub fn random_findex_master_key(
```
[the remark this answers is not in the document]
tbrezot: Yes, we should leave dead code.

Moreover I don't see why you would need to use this function. To my understanding, the master key should be generated by the master authority when a reindexing is done while the token is generated when a user requests it. It seems these are two distinct operations. The master key being generated before a user asks for it.

I don't know when the callback keys should be generated, but I guess it should be something similar.

Plus, you should take `Option`s of your callback keys as input.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+        if keep {
+            Ok(Some(permission))
+        } else {
+            Err(FindexErr::Other(format!(
+                "The token provided doesn't have the permission to {debug_info}"
+            )))
+        }
+    } else {
+        Ok(None)
+    }
```
tbrezot, line 173: It seems this function fails instead of reducing the right. I may be wrong. In any case you should be able to write it in a simpler way using an early return. I think it is equivalent to:
```
if let Some(permission) = permission {
    if !keep {
        return Err(...);
    }
}
Ok(permission)
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+
+enum Callback {
+    FetchEntries,
+    FetchChains,
+    UpsertEntries,
+    InsertChains,
+}
+
+impl FindexCloud {
+    pub fn new(token: String, base_url: Option<String>) -> Result<Self, FindexErr> {
```
tbrezot, line 184: I don't understand why you want to use a string representation of your token. It would be much more compact if you use a byte serialization (~30% less by removing the base64 conversion I think?).

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+            Callback::InsertChains => self.token.insert_chains_key,
+        }
+        .ok_or(FindexErr::Other(format!(
+            "your key doesn't have the permission to call {endpoint}"
+        )))?;
+
+        let mut hasher = Kmac::v128(&key, &[]);
+        let mut output = [0u8; 32];
+        hasher.update(bytes);
+        hasher.finalize(&mut output);
```
tbrezot, line 212: Please use the macro:
```
let output  = kmac!(SIGNATURE_LENGTH, &key, bytes);
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        let signature = hex::encode(output);
+
+        let url = format!(
+            "{}/indexes/{}/{endpoint}",
+            self.base_url
+                .as_deref()
+                .unwrap_or("https://findex.cosmian.com"),
```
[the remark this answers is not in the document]
tbrezot: I agree there may be something to do here. We may not want to hard code this URL.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+        .ok_or(FindexErr::Other(format!(
+            "your key doesn't have the permission to call {endpoint}"
+        )))?;
+
+        let mut hasher = Kmac::v128(&key, &[]);
+        let mut output = [0u8; 32];
+        hasher.update(bytes);
+        hasher.finalize(&mut output);
+
+        let signature = hex::encode(output);
```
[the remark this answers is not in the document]
tbrezot: Agreed. If you can send bytes it would be even better.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+                .as_deref()
+                .unwrap_or("https://findex.cosmian.com"),
+            self.token.index_id,
+        );
+
+        let client = reqwest::Client::new();
+        let res = client
+            .post(url)
+            .header("X-Findex-Cloud-Signature", &signature)
+            .body(bytes.to_vec())
```
tbrezot, line 228: I don't know how post requests work, but maybe you can pass the signature in bytes in the body. Since we know its size, it would be simple to split the body on the receiver's end to recover the signature and the request bytes.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+            .send()
+            .await
+            .map_err(|err| FindexErr::Other(format!("err {err}")))?;
```
tbrezot, line 231: Maybe use a better error message (maybe my review is too early).

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        Ok(res
+            .bytes()
+            .await
+            .map_err(|err| FindexErr::Other(format!("err2 {err}")))?
```
tbrezot, line 236: Same here.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+impl FindexCallbacks<UID_LENGTH> for FindexCloud {
+    async fn progress(
+        &self,
+        _results: &HashMap<Keyword, HashSet<IndexedValue>>,
+    ) -> Result<bool, FindexErr> {
+        Ok(true)
+    }
```
tbrezot, line 247: I agree with @ackRow: the `progress` callback should be removed from the `FindexCallbacks` trait and put back inside the `FindexSearch` trait as it is not managed the same way as the others.

This would complicate this MR so I think you can open an issue about it.

src/interfaces/ffi/api.rs
```diff
@@ -396,3 +300,295 @@ pub unsafe extern "C" fn h_compact(
...
+    indexed_values_and_keywords_ptr: *const c_char,
+) -> c_int {
+    let token = ffi_read_string!("keywords", token_ptr);
+
+    let findex = ffi_unwrap!(FindexCloud::new(token, None));
+    let master_key = findex.token.findex_master_key.clone();
+
+    ffi_upsert(
+        findex,
+        &master_key,
```
tbrezot, line 408: Why do you need to clone this variable?
```suggestion
    ffi_upsert(
        findex,
        &findex.token.findex_master_key,
```

src/interfaces/ffi/api.rs
```diff
@@ -396,3 +300,295 @@ pub unsafe extern "C" fn h_compact(
...
+    fetch_chains_batch_size: c_uint,
+) -> c_int {
+    let token = ffi_read_string!("keywords", token_ptr);
+
+    let findex = ffi_unwrap!(FindexCloud::new(token, None));
+    let master_key = findex.token.findex_master_key.clone();
+
+    ffi_search(
+        findex,
+        &master_key,
```
tbrezot, line 351: Same here, see the next comment

src/interfaces/ffi/api.rs
```diff
@@ -396,3 +300,295 @@ pub unsafe extern "C" fn h_compact(
...
+    max_results_per_keyword: c_int,
+    max_depth: c_int,
+    fetch_chains_batch_size: c_uint,
+) -> c_int {
+    let label_bytes = ffi_read_bytes!("label", label_ptr, label_len);
+    let label = Label::from(label_bytes);
+
+    let max_results_per_keyword = usize::try_from(max_results_per_keyword)
+        .ok()
+        .and_then(NonZeroUsize::new)
```
tbrezot, line 450: why not a `map` ?

src/interfaces/wasm_bindgen/api.rs
```diff
@@ -123,3 +125,147 @@ pub async fn webassembly_upsert(
...
+) -> Result<SearchResults, JsValue> {
+    let mut findex_cloud = FindexCloud::new(token, base_url)?;
+    let master_key = KeyingMaterial::<MASTER_KEY_LENGTH>::try_from_bytes(
+        findex_cloud.token.findex_master_key.as_ref(),
+    )
+    .map_err(|e| JsValue::from(format!("While parsing master key for Findex upsert, {e}")))?;
+
+    let label = Label::from(label_bytes.to_vec());
+
+    let keywords: HashSet<_> = Array::from(&JsValue::from(keywords))
```
tbrezot, line 160: You shouldn't need to explicit the type:
```suggestion
    let keywords = Array::from(&JsValue::from(keywords))
```

src/lib.rs
```diff
@@ -16,6 +16,7 @@
 // It allows the Wasm Findex implementation to reuse the common traits
 // for searching and upserting indexes.
 #![feature(async_fn_in_trait)]
+#![feature(iter_next_chunk)]
```
tbrezot, line 19: I know this is useful, but could we avoid this? It is not good for crypto crates to use unstable features.

Cargo.lock
```diff
@@ -92,7 +92,7 @@ version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
- "hermit-abi",
+ "hermit-abi 0.1.19",
```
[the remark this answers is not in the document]
ThibaudDauce: Let's do that in another PR since it's not related to FindexCloud

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: `random_findex_master_key` is used in wasm and will be used inside FFI next.

We need this function to create the initial token. When creating an index in the console the backend will return all the signing callbacks keys, but the user need to create the token from these key (adding the random findex master key). We could have two operations (one to generate the random findex master key and one to generate the token from all the keys) but it's simpler this way since both operations are done together.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+impl FindexCloud {
+    pub fn new(token: String, base_url: Option<String>) -> Result<Self, FindexErr> {
+        Ok(FindexCloud {
+            token: Token::from_str(&token)?,
+            base_url,
+        })
+    }
+
+    async fn post(&self, callback: Callback, bytes: &[u8]) -> Result<Vec<u8>, FindexErr> {
+        let endpoint = match callback {
```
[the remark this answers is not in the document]
tbrezot: To my understanding each endpoint corresponds the URL information corresponding to its associated callback and each callback has its own key (in the present implementation, this may change to use one key per permission + one key for common callbacks, maybe no key for search).

This gives a one to one endpoint/key mapping but the two are distinct concepts.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+    },
+};
+
+pub(crate) struct FindexCloud {
+    pub(crate) token: Token,
+    pub(crate) base_url: Option<String>,
+}
+
+pub(crate) struct Token {
+    index_id: String,
```
tbrezot, line 32: I don't know what is this `index_id`. What does it represent? Why do we need it? What does it look like?

I understand it is a 4-byte string from the `split` line 74. But that's about all I've got.

Can't you use a fixed-size array of bytes to represent it ?

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: Yes one endpoint = one key.

src/interfaces/ffi/api.rs
ThibaudDauce: 1. `executor::block_on` doesn't allow multi core parallelism when doing requests. I didn't test now, but with `executor::block_on` Java wasn't doing the `fetch_chains` in parallel (massive lost in perfs).
2. `reqwest` require a "real" executor so I switch not only for FindexCloud but because of 1 I switch also the basic Findex to an executor too

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: Why does the user create a random key ? I think we need to keep these two operation separated as I don't understand why we couldn't want to create a token without generating a new random key.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    pub(crate) findex_master_key: KeyingMaterial<MASTER_KEY_LENGTH>,
+
+    fetch_entries_key: Option<[u8; 16]>,
+    fetch_chains_key: Option<[u8; 16]>,
+    upsert_entries_key: Option<[u8; 16]>,
+    insert_chains_key: Option<[u8; 16]>,
+}
+
+impl Display for Token {
```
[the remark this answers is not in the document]
ThibaudDauce: We want to have a string opaque representation for clients to use (to copy paste inside their config / code)

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+                token.fetch_chains_key = key;
+            } else if prefix == 2 {
+                token.upsert_entries_key = key;
+            } else if prefix == 3 {
+                token.insert_chains_key = key;
+            } else {
+                return Err(FindexErr::Other(format!(
+                    "the token contains a unknown prefix {prefix}"
+                )));
+            }
```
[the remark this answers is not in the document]
ThibaudDauce: Cannot do a `for … loop` since I don't know the number of iteration. A token can have less than 4 keys inside it (if it has less permissions). It's not a problem to override an already read key but we could throw an error in this case "duplicate key {key} inside token".

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+            })?);
+
+            if prefix == 0 {
+                token.fetch_entries_key = key;
+            } else if prefix == 1 {
+                token.fetch_chains_key = key;
+            } else if prefix == 2 {
+                token.upsert_entries_key = key;
+            } else if prefix == 3 {
+                token.insert_chains_key = key;
```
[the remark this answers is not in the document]
ThibaudDauce: No because some keys can be absent from the token

src/interfaces/cloud.rs
ThibaudDauce: I switch to a match in 6fdd729, tell me if it's clearer?

src/interfaces/cloud.rs
ThibaudDauce: The token needs to be copy/pastable for a client.

src/interfaces/cloud.rs
ThibaudDauce: I could, but it's not idiomatic of HTTP requests. base64 encoding in header is simpler I think (but we could change this in future if it becomes a problem).

src/interfaces/ffi/api.rs
ThibaudDauce: Because `findex` is move inside the 1rst argument so Rust doesn't allow this… Not a big deal right now, but could be improved for sure at another time.

src/interfaces/ffi/api.rs
ThibaudDauce: Because `NonZeroUsize::new` returns a `Option<>` and we want to flatten `Option<Option<>>` to only `Option<>`.

src/lib.rs
ThibaudDauce: Since we already use nightly for `async_fn_in_trait` I prefer to have the most clean way to achieve de-serialization. (if we were on stable I would have use stable only). It's not really crypto stuff…

src/interfaces/cloud.rs
ThibaudDauce: It's a 5 characters ID representing the index inside the database. I will document the token format.

Cargo.toml
```diff
@@ -38,16 +39,24 @@ js-sys = { version = "0.3", optional = true }
 leb128 = "0.2"
 pyo3 = { version = "0.17", features = ["extension-module", "abi3", "abi3-py37", "generate-import-lib"], optional = true }
 rand = "0.8"
+reqwest = { version = "0.11.14", features = ["rustls-tls"], default-features = false }
```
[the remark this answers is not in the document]
ThibaudDauce: `cargo add` use this format, I'll keep following it and not manually change it. Cargo will use `0.11.15` if available since it follows semver anyway. The only difference with `0.11`: it will just refuse `0.11.13`

Cargo.toml
```diff
@@ -38,16 +39,24 @@ js-sys = { version = "0.3", optional = true }
...
 rusqlite = { version = "0.28", features = ["bundled"], optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
 sha3 = "0.10"
 thiserror = "1"
 tiny-keccak = { version = "2.0.2", features = ["kmac"] }
+tokio = { version = "1.25.0", features = ["rt-multi-thread"], optional = true }
```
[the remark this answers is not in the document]
ThibaudDauce: Same thing as above. I follow `cargo add` and do not manually change version afterward.

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: Because the user need the Findex Master Key for the new index? The backend cannot generate this key since it's a private key.

Right now, we don't want to create a token without generating a new random key, let's keep it simple for clients (the console for exemple) with only one call to wasm, if we need to separate in the future we could create the two distinct operations and keep this one as an alias.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        write!(f, "{}{}", self.index_id, base64::encode(keys))
+    }
+}
+
+impl FromStr for Token {
+    type Err = FindexErr;
+
+    fn from_str(token: &str) -> Result<Self, Self::Err> {
+        let (index_id, tail) = token.split_at(5);
```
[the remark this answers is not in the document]
ThibaudDauce: It's the format chosen by me :-) The index ID is 5 characters long (to avoid too much collision but not take a lot of space inside the end-user token).

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: Yes I think we could do that.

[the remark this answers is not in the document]
tbrezot: Or maybe you could check the size beforehand. I am bothered by the `while` loop. But maybe it's just me.

src/interfaces/cloud.rs
tbrezot: Yes it is much better

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: Then it needs to be clearly specified, using a constant and some doc.

src/lib.rs
tbrezot: Well it is inside a cryptographic repo. It's okay because I will move it to CloudproofRust later. But I still don't really like it :)

src/interfaces/ffi/api.rs
tbrezot: True, I didn't see it. Why do you need the token to be part of Findex?

src/interfaces/cloud.rs
tbrezot: ok

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: ok

Cargo.toml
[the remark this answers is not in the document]
Manuthor: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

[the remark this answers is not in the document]
tbrezot: Actually specifying all version numbers blocks minor upgrades *only* for crates in `0.x.x` which are still unstable, which is a good thing. For crates in `X.x.x` with `X>0` specifying all version numbers *allows* minor upgrades.

Therefore, I think we can start specifying dependencies using all version numbers in all our crates.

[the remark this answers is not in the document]
tbrezot: It will be moved to CloudproofRust later on but could you please make this dependency optional?

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+impl Token {
+    pub fn random_findex_master_key(
+        index_id: String,
+        fetch_entries_key: [u8; 16],
+        fetch_chains_key: [u8; 16],
+        upsert_entries_key: [u8; 16],
+        insert_chains_key: [u8; 16],
```
tbrezot, line 132: Could you take optional parameters here?

I understand you prefer to create a full object in one client call, but the wrapping should be done at the interface level I think.

src/interfaces/cloud.rs
tbrezot: Shouldn't this be managed on a higher level ?

tbrezot: If you do it client side, you would save allocation space and bandwidth.

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: Actually, I think it should be cleaner to generate the master key in the interface code. The token would still be initialized with a new random key upon a unique wasm call but the Rust code shouldn't mix these concepts imho.

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: I think Findex cloud should only manage the serialization part. The String representation should be left to the client.

Cargo.toml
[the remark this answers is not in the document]
ThibaudDauce: Done in 4155642

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: It's hard to have a correct random generator in all languages (for example in JS `crypto` is different in NodeJS and browser so it require some hard packaging to allow both target, having it in Rust is simpler)

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: It's not really hard-coded since it's only the default value?

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: I don't think so, having the string representation in the client do not allow for simple sharing of the token between clients (each client have it's own way to represent the string…). I think it's better that the Rust is responsible for the opaque string token (I know you don't like base64/hex inside Rust :-D )

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: Yes

src/interfaces/cloud.rs
ThibaudDauce: Do you know if there is a way to call the macro with `let output = kmac!(Self::SIGNATURE_LENGTH, &key, bytes);`

```
   --> src/interfaces/cloud.rs:214:32
    |
214 |         let output = kmac!(Self::SIGNATURE_LENGTH, &key, bytes);
    |                                ^^ no rules expected this token in macro call
```

src/interfaces/cloud.rs
ThibaudDauce: Good catch, fixed in c2663b7 :-)

src/interfaces/cloud.rs
ThibaudDauce: Good catch, fixed in c2663b7 :-)

src/interfaces/ffi/api.rs
ThibaudDauce: Because the token contains the keys and the keys are required in the callbacks.

src/interfaces/wasm_bindgen/api.rs
ThibaudDauce: Good catch, fixed in 0be88d8

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: I mean you could still generate the randomness in Rust, but inside the WASM interface (for example). This would give something like:
wasm_call -------> rust_wasm_interface --------------> generate random key
                                                                     |
                                                                      --------------> generate token

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: True, but why is this value the default one ? Are we sure we will not change it?

I actually think it might be safer to remove the option at once.

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: I am pretty sure there exist a standard fr the base64 conversion. Every client should be able to correctly convert a base64 string into bytes (we already rely on this in Findex interfaces). The string representation would only be needed for display (for the copy/paste) and an array of bytes should be used internally by the clients.

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: We're building a SAAS offering, it'll have a default domain (maybe not this one but it could be). Custom domain is only supported for development purposes (changing it to https://findex.staging.cosmian.com) or for on premise deployment in the future. 

Having the default value in Rust avoid copy pasting it in all clients, if we want to change it (for exemple for a Findex v2) we have a single source of truth (but it should be a constant)

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: That way:
- we keep a unique representation everywhere (the byte serialization)
- we keep the code used for conversion (serialization) in one place (Rust) (we use base64 libraries on client side for display only)
- clients still can copy/paste easily
- we save time and space in callbacks

src/interfaces/cloud.rs
tbrezot: Indeed, the macros asks for an `ident`... you can replace it by `expr` in the macro definition to make it work.

Otherwise you would need to create a local copy of the length:
```
let SIGNATURE_LENGTH: usize = Self::SIGNATURE_LENGTH;
let output = kmac!(SIGNATURE_LENGTH, &key, bytes);
```

This should work and is not too expensive.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,345 @@
...
+        ser_de::serialize_set,
+    },
+};
+
+pub(crate) struct FindexCloud {
+    pub(crate) token: Token,
+    pub(crate) base_url: Option<String>,
+}
+
+pub(crate) struct Token {
```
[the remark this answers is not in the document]
ThibaudDauce: Is it good in ff89a2a?

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: I don't want to have copy/paste inside FFI/Wasm if it's the same in both, no?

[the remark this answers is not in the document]
tbrezot: This is just two lines. I would do it, but I understand your point. Maybe we can ask @Manuthor to give us his opinion?

src/interfaces/cloud.rs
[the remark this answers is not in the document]
tbrezot: Okay, up to you :)

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: @Manuthor is it clearer with this commit? 0aba99b

src/interfaces/cloud.rs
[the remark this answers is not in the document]
Manuthor: ok good for me

src/interfaces/wasm_bindgen/api.rs
ThibaudDauce: good catch, fixed in 8a0f6f0

src/interfaces/ffi/api.rs
ThibaudDauce: good catch, fixed in 8a0f6f0

src/interfaces/cloud.rs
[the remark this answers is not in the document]
ThibaudDauce: Is it better for you inside ff89a2a?

src/interfaces/cloud.rs
ThibaudDauce: Done in 0fa0fa7

src/interfaces/cloud.rs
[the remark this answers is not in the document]
Manuthor: I would keep the random generation in Rust (can be tricky to have good implementations in every language).

I agree that the 2 operations should be done separately (master key generation and authentication keys generation). But, for convenience, I agree that this token should be also self-sufficient (and then generated in one pass).

But one question: since master key is renewed at each call, how do you deliver multiple tokens with the same Findex master key?

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,391 @@
...
+///     2. 1 byte prefix identifying the next key
+///     3. 16 bytes of callback signature key
+///     4. 1 byte prefix identifying the next key
+///     5. …
+///
+/// Currently each callback have an associated signature key used in a kmac to
+/// send request to the backend. These key are only used for authorization
+/// and do not secure the index (the findex master key do). In the future, we
+/// could do optimization to avoid having one key for each callback but we want
+/// to disallow the server to differenciate a `fetch_entries` for a search or a
```
Manuthor, line 61:
```suggestion
/// to disallow the server to differentiate a `fetch_entries` for a search or a
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+/// This size allows us to have a lot of indexes for ou backend while
+/// obfuscating the number of index present (contrary to incremented ID).
+/// Moreover, 5 is a small enough value to have inside the size-limited token
+/// (contrary to UUID).
+pub const PUBLIC_INDEX_ID_LENGTH: usize = 5;
+
+/// The callback signature is a kmac of the body of the request used to do
+/// autorization (checking if this client can call this callback)
```
Manuthor, line 38:
```suggestion
/// authorization (checking if this client can call this callback)
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,391 @@
...
+                    "Impossibe to send the request to FindexCloud: {err}"
+                ))
+            })?;
+
+        Ok(res
+            .bytes()
+            .await
+            .map_err(|err| {
+                FindexErr::Other(format!(
+                    "Impossibe to read the returned bytes from FindexCloud:  {err}"
```
Manuthor, line 280:
```suggestion
                    "Impossible to read the returned bytes from FindexCloud:  {err}"
```

src/interfaces/cloud.rs
ThibaudDauce: @Manuthor Is 566f115 more clear?

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,408 @@
...
+/// and do not secure the index (the findex master key do). In the future, we
+/// could do optimization to avoid having one key for each callback but we want
+/// to disallow the server to differentiate a `fetch_entries` for a search or a
+/// `fetch_entries` for an upsert while still allowing fine grain permissions.
+pub(crate) struct Token {
+    /// This ID identify an index inside the Findex Cloud backend
+    /// We do not use auto-increment integer ID because we don't want to leak
+    /// the number of indexes inside our database.
+    /// We do not use UUID because the token is limited in space.
+    /// The abritrary chosen length is `INDEX_ID_LENGTH`.
```
Manuthor, line 68:
```suggestion
    /// The arbritrary chosen length is `INDEX_ID_LENGTH`.
```

src/interfaces/cloud.rs
Manuthor: thanks for the additional comments.
Is this `index_id` sensitive? is it related to the Findex algorithm? Isn't it a problem this size limit on 5 characters?

Manuthor: another remark: why can't we have an another master key (an authentication_master_key) which will be derived in 4 authentication keys? (it will reduce the token size)

ThibaudDauce: `index_id` is not sensitive, it can be public it only serve to identify the index and obfuscate the number of index inside our database. See my new comment in 31f6a39

ThibaudDauce: > another remark: why can't we have an another master key (an authentication_master_key) which will be derived in 4 authentication keys? (it will reduce the token size)

I think I already propose something similar to @chloehebant but it wasn't good security wise… But what you propose is a little bit different, maybe it's fine? I will ask her.

ThibaudDauce: > > another remark: why can't we have an another master key (an authentication_master_key) which will be derived in 4 authentication keys? (it will reduce the token size)
> 
> I think I already propose something similar to @chloehebant but it wasn't good security wise… But what you propose is a little bit different, maybe it's fine? I will ask her.

It's not working because with this system we cannot authorize a client to do one callback and not another?

tbrezot: Indeed, if we have only one key used to derive the others, there is no way to prevent a given client from deriving any callback authentication key.

tbrezot: The maximum we can do to preserve rights is to use one seed (what we call "key" in this discussion is a "seed" from a cryptographic point of view since it is used to derive other keys) per right. A seed can then be used to derive the callback authorization keys associated to the corresponding right.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+    str::FromStr,
+};
+
+use cosmian_crypto_core::{bytes_ser_de::Serializable, CsRng};
+use rand::SeedableRng;
```
tbrezot, line 8: Please, use `cosmian_crypto_core::reexport::rand_core::SeedableRng`. It would be great to be able to get rid of the `rand` dependency.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+pub(crate) struct FindexCloud {
+    pub(crate) token: Token,
+    pub(crate) base_url: Option<String>,
+}
+
+/// See `Token@index_id`
+pub const INDEX_ID_LENGTH: usize = 5;
+
+/// The callback signature is a kmac of the body of the request used to do
+/// authorization (checking if this client can call this callback)
```
tbrezot, line 35:
```suggestion
/// The callback signature is a kmac of the body of the request.
/// It is used to assert the client can call this callback.
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+
+/// Findex Cloud tokens are a string containing all information required to do
+/// requests to Findex Cloud (except the label because it is a value changing a
+/// lot).
+///
+/// The string is encoded as follow:
+/// 1. `index_id` `INDEX_ID_LENGTH` chars (see `Token@index_id`)
+/// 2. base64 representation of the different keys:
+///     1. `SIGNATURE_KEY_LENGTH` bytes of findex master key (this key is never
+/// sent to the Findex Cloud backend)
```
tbrezot, line 51:
```suggestion
///     1. `MASTER_KEY_LENGTH` bytes of findex master key (this key is never
/// sent to the Findex Cloud backend)
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+///     2. 1 byte prefix identifying the next key
+///     3. `SIGNATURE_KEY_LENGTH` bytes of callback signature key
+///     4. 1 byte prefix identifying the next key
+///     5. …
+///
+/// Currently each callback have an associated signature key used in a kmac to
```
tbrezot, line 57:
```suggestion
/// Currently each callback has an associated signature key used in a kmac to
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+/// send request to the backend. These key are only used for authorization
+/// and do not secure the index (the findex master key do). In the future, we
+/// could do optimization to avoid having one key for each callback but we want
+/// to disallow the server to differentiate a `fetch_entries` for a search or a
+/// `fetch_entries` for an upsert while still allowing fine grain permissions.
+pub(crate) struct Token {
+    /// This ID identify an index inside the Findex Cloud backend
```
tbrezot, line 64:
```suggestion
    /// This ID identifies an index inside the Findex Cloud backend
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// This number is not sensitive, it's only an ID. If someone find this ID,
```
tbrezot, line 65:
```suggestion
    /// This number is not sensitive, it's only an ID. If someone finds this ID,
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    fetch_entries_key: Option<KeyingMaterial<SIGNATURE_KEY_LENGTH>>,
+    fetch_chains_key: Option<KeyingMaterial<SIGNATURE_KEY_LENGTH>>,
+    upsert_entries_key: Option<KeyingMaterial<SIGNATURE_KEY_LENGTH>>,
+    insert_chains_key: Option<KeyingMaterial<SIGNATURE_KEY_LENGTH>>,
+}
+
+impl Display for Token {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut keys = self.findex_master_key.to_vec();
```
tbrezot, line 83: Since we know how many bytes we are going to write, you could allocate the required space at once.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+
+impl FromStr for Token {
+    type Err = FindexErr;
+
+    fn from_str(token: &str) -> Result<Self, Self::Err> {
+        let (index_id, tail) = token.split_at(INDEX_ID_LENGTH);
+        let mut bytes = base64::decode(tail)
+            .map_err(|_| {
+                FindexErr::Other(format!(
+                    "token {token} is not a valid base64 encoded string"
```
tbrezot, line 117: I am a bit skeptical about this assertion since I am not sure the token *is* a valid base64 string. Only `tail` should be. I am pretty sure token is actually *not* a valid base64 string since `INDEX_ID_LENGTH` is 5 which makes the token length an odd number.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+        };
+
+        while let Some(prefix) = bytes.next() {
+            let key = Some(
+                bytes
+                    .next_chunk::<SIGNATURE_KEY_LENGTH>()
+                    .map_err(|_| {
+                        FindexErr::Other(format!(
+                            "the token is too short, expecting {SIGNATURE_KEY_LENGTH} bytes after \
+                             the prefix {prefix}"
```
tbrezot, line 147: Maybe return the offset in the token as the `prefix` may not be easy to locate (this is just a byte among other bytes which may have the same value).

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+                token.fetch_entries_key = key;
+            } else if prefix == 1 {
+                token.fetch_chains_key = key;
+            } else if prefix == 2 {
+                token.upsert_entries_key = key;
+            } else if prefix == 3 {
+                token.insert_chains_key = key;
+            } else {
+                return Err(FindexErr::Other(format!(
+                    "the token contains a unknown prefix {prefix}"
```
tbrezot, line 163: Same here, maybe you can give the offset.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+        &self,
+        token: &'a Token,
+    ) -> &'a Option<KeyingMaterial<SIGNATURE_KEY_LENGTH>> {
+        match self {
+            Callback::FetchEntries => &token.fetch_entries_key,
+            Callback::FetchChains => &token.fetch_chains_key,
+            Callback::UpsertEntries => &token.upsert_entries_key,
+            Callback::InsertChains => &token.insert_chains_key,
+        }
+    }
```
tbrezot, line 249: Maybe you could move this function in the `Token` implementation.

Wouldn't this be the right place to derive the actual keys  ?

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+}
+
+impl FindexCloud {
+    pub fn new(token: String, base_url: Option<String>) -> Result<Self, FindexErr> {
```
tbrezot, line 253: You don't need the ownership here.
```suggestion
    pub fn new(token: &str, base_url: Option<String>) -> Result<Self, FindexErr> {
```

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    async fn post(&self, callback: Callback, bytes: &[u8]) -> Result<Vec<u8>, FindexErr> {
+        let endpoint = callback.get_uri();
+
+        let key: KmacKey = callback
+            .get_key(&self.token)
+            .as_ref()
+            .ok_or(FindexErr::Other(format!(
+                "your key doesn't have the permission to call {endpoint}"
```
tbrezot, line 267: There seem to be a confusion between `endpoint` and the callback string representation. Maybe implement `Display` for `Callback` and use it here? You can still use it in `get_uri()` if you want to use the same string as endpoint.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+            )))?
+            .derive_kmac_key(self.token.index_id.as_bytes());
```
tbrezot, line 269: See my previous comment: it may be better if this is abstracted away by the `get_key()` method.

src/interfaces/cloud.rs
```diff
@@ -0,0 +1,410 @@
...
+                .as_deref()
+                .unwrap_or(FINDEX_CLOUD_DEFAULT_DOMAIN),
+            self.token.index_id,
+        );
+
+        let client = reqwest::Client::new();
+        let res = client
+            .post(url)
+            .header("X-Findex-Cloud-Signature", &signature)
+            .body(bytes.to_vec())
```
tbrezot, line 285: Do you think it would be a good idea to take ownership of `bytes`? I always prefer when there is no hidden clone from a caller point of view.

src/interfaces/cloud.rs
tbrezot: Done in #24

src/interfaces/cloud.rs
ThibaudDauce: Done in f00c141

src/interfaces/cloud.rs
ThibaudDauce: It is a lot of code, do you have an idea how to simplify it?

```
        let size = self.findex_master_key.len()
            + self
                .fetch_entries_key
                .as_ref()
                .map(|key| key.len() + 1)
                .unwrap_or(0)
            + self
                .fetch_chains_key
                .as_ref()
                .map(|key| key.len() + 1)
                .unwrap_or(0)
            + self
                .upsert_entries_key
                .as_ref()
                .map(|key| key.len() + 1)
                .unwrap_or(0)
            + self
                .insert_chains_key
                .as_ref()
                .map(|key| key.len() + 1)
                .unwrap_or(0);
```

src/interfaces/cloud.rs
ThibaudDauce: Changed the wording in 424a022, is it ok?

src/interfaces/cloud.rs
ThibaudDauce: Done in 424a022

src/interfaces/cloud.rs
ThibaudDauce: Done in 424a022

src/interfaces/cloud.rs
ThibaudDauce: Yes good idea, simplify the code a lot! d47b268

src/interfaces/cloud.rs
ThibaudDauce: Done in d47b268

src/interfaces/cloud.rs
ThibaudDauce: Yes good idea, done in d47b268

src/interfaces/cloud.rs
ThibaudDauce: I think there is no real sense to generate a new token with not all permissions. This operation is done before the usage of the index (since before this call there is no Findex Master key) and the user calling this function should have all the permissions on his index (since he is the one generating the Findex Master Key). Do you agree?

src/interfaces/cloud.rs
tbrezot: Yes it is better, maybe just change `the keys section are` into `the key section is`?

Manuthor (approved):

tbrezot (approved):

Manuthor (approved):

PR #22, opened 2023-02-08
Feature/cloudproof rust

tbrezot:


findex_in_memory/src/findex.rs
```diff
@@ -0,0 +1,248 @@
...
+        BLOCK_LENGTH,
+        TABLE_WIDTH,
+        MASTER_KEY_LENGTH,
+        KWI_LENGTH,
+        KMAC_KEY_LENGTH,
+        DEM_KEY_LENGTH,
+        KmacKey,
+        DemScheme,
+        Error,
+    > for FindexInMemory<UID_LENGTH>
```
ThibaudDauce, line 230: We should create a type alias for this… Maybe not in this PR…

findex_sqlite/src/error.rs
```diff
@@ -0,0 +1,53 @@
+use std::fmt::Display;
+
+#[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
```
ThibaudDauce, line 4: Could we comply to clippy here instead of ignoring? It's not a big change?

findex_sqlite/src/ser_de.rs
```diff
@@ -56,7 +53,7 @@ where
     }
 
     fn read(_de: &mut Deserializer) -> Result<Self, Self::Error> {
-        Err(FindexErr::Other(
+        Err(Error::CryptoError(
```
ThibaudDauce, line 56: Is it a crypto error here? Doesn't look like one…

ThibaudDauce (approved):

findex_sqlite/src/error.rs
tbrezot: Could you explain yourself ?

findex_in_memory/src/findex.rs
tbrezot: I agree, but I prefer this to be done in a latter PR :pray:

findex_sqlite/src/ser_de.rs
tbrezot: Indeed, a `CoversionError` seems more appropriate, thanks

findex_in_memory/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
```
Manuthor, line 1: if you don't mind, we could move folders `findex_in_memory` and `findex_sqlite` to the root folder `examples` (in order to be build via `cargo build --examples`)

tbrezot: I tried several possibilities but the only one allowing me to do what I wanted is the present one...

What I want is to be able to import structures defined in these crates for tests and benchmarks (at least). This is not possible (at least I couldn't get it working) by putting them in the `tests` directory. I bet it would be the same with the `examples` directory.

findex_in_memory/src/findex.rs
tbrezot: Done in #25

findex_sqlite/src/ser_de.rs
tbrezot: Done in 578e60e

findex_sqlite/src/error.rs
ThibaudDauce: Clippy is complaining that their is an `Error` suffix, no? Could we remove the `Error` suffix inside the enum instead of `#[allow(clippy::enum_variant_names)]`?

tbrezot: I removed it, clippy doesn't seem to comply anymore...

tbrezot: In eaf2174

PR #31, opened 2023-03-06
Support deletions

tbrezot:
Support deleting an indexed value from the list of indexed value of a given keyword.

- the upsert interface now takes as input two distinct lists for the additions and the deletions.
  **Question**: an alternative would be to flag each element of a single list as being an addition or a deletion. I don't know which is the best.
- the Chain Table size is increased by one byte per line.
- the benches have been **improved** (an effort has been made to reduce the overall number of allocations).

Below are given the benches with a comparison to the `feature/add_macros` branch.
```
search/Searching 1 word time:   [10.643 ms 10.676 ms 10.709 ms]
                        change: [-11.934% -11.422% -10.879%] (p = 0.00 < 0.05)
                        Performance has improved.
search/Searching 10 words
                        time:   [10.802 ms 10.832 ms 10.864 ms]
                        change: [-10.606% -10.073% -9.5247%] (p = 0.00 < 0.05)
search/Searching 100 words
                        time:   [11.752 ms 11.790 ms 11.828 ms]
                        change: [-12.358% -11.833% -11.286%] (p = 0.00 < 0.05)
search/Searching 1000 words
                        time:   [21.582 ms 21.630 ms 21.679 ms]
                        change: [-3.1334% -2.7810% -2.4431%] (p = 0.00 < 0.05)
upsert/Indexing 20 keywords
                        time:   [234.91 µs 235.71 µs 236.55 µs]
                        change: [-2.5067% -1.9320% -1.3567%] (p = 0.00 < 0.05)
upsert/Indexing 200 keywords
                        time:   [2.2661 ms 2.2743 ms 2.2828 ms]
                        change: [-8.9163% -8.3586% -7.8004%] (p = 0.00 < 0.05)
upsert/Indexing 2000 keywords
                        time:   [24.211 ms 24.287 ms 24.364 ms]
                        change: [-2.8158% -2.3230% -1.8367%] (p = 0.00 < 0.05)
```

### DIFF
--- a/src/chain_table.rs
+++ b/src/chain_table.rs
@@ -283,8 +283,7 @@ mod tests {
         let indexed_value_1 = IndexedValue::from(Location::from("location1".as_bytes()));
         let indexed_value_2 = IndexedValue::from(Location::from("location2".as_bytes()));
         let mut chain_table_value = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in
-            Block::<BLOCK_LENGTH>::pad(InsertionType::Addition, &indexed_value_1.to_vec()).unwrap()
+        for block in Block::<BLOCK_LENGTH>::pad(InsertionType::Addition, &indexed_value_1).unwrap()
         {
             chain_table_value.try_push(block).unwrap();
         }
@@ -292,8 +291,7 @@ mod tests {
         assert_eq!(chain_table_value.length(), bytes.len());
         let res = ChainTableValue::try_from_bytes(&bytes).unwrap();
         assert_eq!(chain_table_value, res);
-        for block in
-            Block::<BLOCK_LENGTH>::pad(InsertionType::Addition, &indexed_value_2.to_vec()).unwrap()
+        for block in Block::<BLOCK_LENGTH>::pad(InsertionType::Addition, &indexed_value_2).unwrap()
         {
             chain_table_value.try_push(block).unwrap();
         }
@@ -316,17 +314,17 @@ mod tests {
         let indexed_value2 = IndexedValue::from(location);
 
         let mut chain_table_value1 = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in Block::pad(InsertionType::Addition, &indexed_value1.to_vec()).unwrap() {
+        for block in Block::pad(InsertionType::Addition, &indexed_value1).unwrap() {
             chain_table_value1.try_push(block).unwrap();
         }
-        for block in Block::pad(InsertionType::Deletion, &indexed_value2.to_vec()).unwrap() {
+        for block in Block::pad(InsertionType::Deletion, &indexed_value2).unwrap() {
             chain_table_value1.try_push(block).unwrap();
         }
         // The indexed values should be short enough to fit in a single block.
         assert_eq!(chain_table_value1.length, 2);
 
         let mut chain_table_value2 = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in Block::pad(InsertionType::Addition, &indexed_value1.to_vec()).unwrap() {
+        for block in Block::pad(InsertionType::Addition, &indexed_value1).unwrap() {
             chain_table_value2.try_push(block).unwrap();
         }
         // The indexed values should be short enough to fit in a single block.
--- a/src/chain_table.rs
+++ b/src/chain_table.rs
@@ -283,14 +283,14 @@ mod tests {
         let indexed_value_1 = IndexedValue::from(Location::from("location1".as_bytes()));
         let indexed_value_2 = IndexedValue::from(Location::from("location2".as_bytes()));
         let mut chain_table_value = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in Block::<BLOCK_LENGTH>::pad(BlockType::Addition, &indexed_value_1).unwrap() {
+        for block in indexed_value_1.to_blocks(BlockType::Addition).unwrap() {
             chain_table_value.try_push(block).unwrap();
         }
         let bytes = chain_table_value.try_to_bytes().unwrap();
         assert_eq!(chain_table_value.length(), bytes.len());
         let res = ChainTableValue::try_from_bytes(&bytes).unwrap();
         assert_eq!(chain_table_value, res);
-        for block in Block::<BLOCK_LENGTH>::pad(BlockType::Addition, &indexed_value_2).unwrap() {
+        for block in indexed_value_2.to_blocks(BlockType::Addition).unwrap() {
             chain_table_value.try_push(block).unwrap();
         }
         let bytes = chain_table_value.try_to_bytes().unwrap();
@@ -312,17 +312,17 @@ mod tests {
         let indexed_value2 = IndexedValue::from(location);
 
         let mut chain_table_value1 = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in Block::pad(BlockType::Addition, &indexed_value1).unwrap() {
+        for block in indexed_value1.to_blocks(BlockType::Addition).unwrap() {
             chain_table_value1.try_push(block).unwrap();
         }
-        for block in Block::pad(BlockType::Deletion, &indexed_value2).unwrap() {
+        for block in indexed_value2.to_blocks(BlockType::Deletion).unwrap() {
             chain_table_value1.try_push(block).unwrap();
         }
         // The indexed values should be short enough to fit in a single block.
         assert_eq!(chain_table_value1.length, 2);
 
         let mut chain_table_value2 = ChainTableValue::<CHAIN_TABLE_WIDTH, BLOCK_LENGTH>::default();
-        for block in Block::pad(BlockType::Addition, &indexed_value1).unwrap() {
+        for block in indexed_value1.to_blocks(BlockType::Addition).unwrap() {
             chain_table_value2.try_push(block).unwrap();
         }
         // The indexed values should be short enough to fit in a single block.
--- a/src/chain_table.rs
+++ b/src/chain_table.rs
@@ -21,124 +21,151 @@ use std::{
 };
 
 use cosmian_crypto_core::{
-    bytes_ser_de::{Deserializer, Serializer},
+    bytes_ser_de::{Deserializer, Serializable, Serializer},
     reexport::rand_core::CryptoRngCore,
     symmetric_crypto::{Dem, SymKey},
 };
 
 use crate::{
     error::CoreError as Error,
-    structs::{Block, Uid},
+    structs::{Block, BlockType, Uid},
     KeyingMaterial, CHAIN_TABLE_KEY_DERIVATION_INFO,
 };
 
 /// Value of the Chain Table. It is composed of a maximum of `TABLE_WIDTH`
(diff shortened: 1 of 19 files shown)


tests/test_in_memory.rs
```diff
@@ -415,6 +425,36 @@ async fn test_findex() -> Result<(), Error<ExampleError>> {
...
+        HashSet::from_iter(vec![doe_keyword.clone()]),
+    );
+    findex
+        .upsert(
+            HashMap::new(),
+            indexed_value_to_keywords,
+            &master_key,
+            &new_label,
+        )
+        .await?;
```
ackRow, line 441: Maybe rename `indexed_value_to_keywords` to `indexed_value_to_keywords_removal`?

src/chain_table.rs
```diff
@@ -21,121 +21,154 @@ use std::{
...
-            res.push(block);
-        }
-        Ok(Self(res))
+        Self::try_from_bytes(&bytes)
     }
 }
 
-impl<const BLOCK_LENGTH: usize> Deref for ChainTableValue<BLOCK_LENGTH> {
-    type Target = Vec<Block<BLOCK_LENGTH>>;
+impl<const TABLE_WIDTH: usize, const BLOCK_LENGTH: usize> Serializable
```
Manuthor, line 114: No remark about this serialization (LGTM). But could you please update the README.md in order to detail the format of the 2 `index` tables. Ideally the structure (including the underlying structure of a `block` and the padding used) and the size of the chain table value should be given.

Also could you please add a table with some very basic examples of the expected size of these indexes (without graphs) when: 
- indexing 1 keyword gives an overhead of ... 
- indexing 2 keywords gives an overhead of ...
- ...

Something straightforward:

| Nb of words indexed | Nb of Entry Table Items | Nb of Chain Table Items | Size of Entry Table Item (in bytes) | Size of Chain Table Item (in bytes) | 
|---------------------|-------------------------|------------------------|----------------------------------|----------------------------------|
| 1                   | 1                       | 1                       | ...                      | ...                              | ...                 |
| 2                   | ...                       | ...                       | ...                      | ...                              | ...                 |

Manuthor: do you think we could have a security reason why we have chosen TABLE_WIDTH = 5 and BLOCK_NUMBER=32 for the findex interfaces?

src/chain_table.rs
```diff
@@ -21,121 +21,154 @@ use std::{
...
 };
 
 /// Value of the Chain Table. It is composed of a maximum of `TABLE_WIDTH`
 /// blocks of length `BLOCK_LENGTH`.
 #[must_use]
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
-pub struct ChainTableValue<const BLOCK_LENGTH: usize>(Vec<Block<BLOCK_LENGTH>>);
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ChainTableValue<const TABLE_WIDTH: usize, const BLOCK_LENGTH: usize> {
+    length: usize,
```
Manuthor, line 40: what does `length` represent here? the number of blocks?

tbrezot: Yes, since I switched for a fixed-size representation of the `ChainTableValue`, I need a way to keep track of the number of blocks that are not padding. This is not part of the serialization so no overhead is added, but having a fixed-size object allows for some performance improvements.

src/chain_table.rs
tbrezot: done in 5fb568e

src/chain_table.rs
```diff
@@ -21,121 +21,154 @@ use std::{
...
-        const TABLE_WIDTH: usize,
-        const DEM_KEY_LENGTH: usize,
-        DEM: Dem<DEM_KEY_LENGTH>,
-    >(
+    pub fn decrypt<const DEM_KEY_LENGTH: usize, DEM: Dem<DEM_KEY_LENGTH>>(
         kwi_value: &DEM::Key,
         ciphertext: &[u8],
     ) -> Result<Self, Error> {
-        if TABLE_WIDTH * BLOCK_LENGTH + DEM::ENCRYPTION_OVERHEAD != ciphertext.len() {
+        if 1 + TABLE_WIDTH * (1 + BLOCK_LENGTH) + DEM::ENCRYPTION_OVERHEAD != ciphertext.len() {
```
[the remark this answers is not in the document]
tbrezot: done in 5fb568e

tests/test_in_memory.rs
tbrezot: I changed to `deletions`.

ackRow (approved):

Manuthor (approved):

src/chain_table.rs
```diff
@@ -21,121 +21,167 @@ use std::{
...
-
-impl<const BLOCK_LENGTH: usize> IntoIterator for ChainTableValue<BLOCK_LENGTH> {
-    type IntoIter = std::vec::IntoIter<Block<BLOCK_LENGTH>>;
-    type Item = Block<BLOCK_LENGTH>;
 
-    fn into_iter(self) -> Self::IntoIter {
-        self.0.into_iter()
+    fn read(de: &mut Deserializer) -> Result<Self, Self::Error> {
+        let mut res = Self::default();
+        let mut flag = de.read_array::<1>()?[0];
```
ThibaudDauce, line 155: Flag could be not mutable here by using `flag & (1 << i)` to check if the bit is set (or something similar, doing it in my head ^^)

src/compact.rs
```diff
@@ -139,12 +140,11 @@ pub trait FindexCompact<
...
         for (kwi, chain) in chains_to_reindex {
-            let mut indexed_values = HashSet::new();
-            let blocks = chain.into_iter().flat_map(|(_, v)| v).collect::<Vec<_>>();
-            for bytes in Block::unpad(&blocks)? {
-                indexed_values.insert(IndexedValue::try_from_bytes(&bytes)?);
-            }
-            reindexed_chain_values.insert(kwi.clone(), indexed_values);
+            let blocks = chain
+                .into_iter()
+                .flat_map(|(_, chain_value)| chain_value.as_blocks().to_vec())
```
ThibaudDauce, line 145: Too bad we require an allocation here with the `to_vec()`… Can we do without it?

src/search.rs
```diff
@@ -127,11 +125,11 @@ pub trait FindexSearch<
...
-            let entry = res.entry(keyword.clone()).or_default();
-            let blocks = chain.into_iter().flat_map(|(_, v)| v).collect::<Vec<_>>();
-            for bytes in Block::unpad(&blocks)? {
-                entry.insert(IndexedValue::try_from_bytes(&bytes)?);
-            }
+            let blocks = chain
+                .into_iter()
+                .flat_map(|(_, chain_table_value)| chain_table_value.as_blocks().to_vec())
+                .collect::<Vec<_>>();
+            res.insert(keyword.clone(), IndexedValue::from_blocks(blocks)?);
```
ThibaudDauce, line 132: Same here, could `IndexedValue::from_blocks` take an iterator or something similar to avoid unnecessary copy?

src/structs.rs
```diff
@@ -61,132 +61,120 @@ impl Keyword {
...
+impl From<u8> for BlockPrefix {
+    fn from(value: u8) -> Self {
+        if 0 == value {
+            Self::Padding
+        } else if u8::MAX == value {
+            Self::NonTerminating
+        } else {
+            Self::Terminating { length: value }
+        }
+    }
```
ThibaudDauce, line 96: Maybe use a match here to have symmetry with the inverse operation above?

src/structs.rs
```diff
@@ -61,132 +61,120 @@ impl Keyword {
...
+pub struct Block<const LENGTH: usize> {
+    pub(crate) block_type: BlockType,
+    pub(crate) prefix: BlockPrefix,
+    pub(crate) data: [u8; LENGTH],
+}
 
-impl_byte_array!(Block);
+/// A block length should be a valid `u8` to allow writing in one byte, and the
+/// value `u8::MAX` is reserved to indicate a non-terminating block.
+const MAX_BLOCK_LENGTH: usize = 254; // `u8::MAX - 1`
```
ThibaudDauce, line 132:
```suggestion
const MAX_BLOCK_LENGTH: usize = (u8::MAX - 1) as usize;
```

src/structs.rs
```diff
@@ -61,132 +61,120 @@ impl Keyword {
...
 
 impl<const LENGTH: usize> Block<LENGTH> {
+    /// This checks that the `BLOCK_LENGTH` does not exceed the
+    /// `MAX_BLOCK_LENGTH`.
+    pub const CHECK_LENGTH: () = assert!(
+        LENGTH <= MAX_BLOCK_LENGTH,
+        "`BLOCK_LENGTH` should be *not* be greater than 254",
```
ThibaudDauce, line 139: Do not hardcode 254 here?

src/structs.rs
```diff
@@ -61,132 +61,120 @@ impl Keyword {
...
+    /// Returns `true` if this block is an addition, and `false` if it is a
+    /// deletion.
+    pub fn is_addition(&self) -> bool {
+        self.block_type == BlockType::Addition
+    }
+
+    /// Generates a new padding block.
+    pub const fn padding_block() -> Self {
+        Self {
+            block_type: BlockType::Deletion,
```
ThibaudDauce, line 174: Why padding blocks are `BlockType::Deletion`? In my understanding it doesn't matter but :
1. Is it hard to have `block_type` info only on not padding block?
2. If it is, can you add a comment explaining that this value is meaningless?

src/compact.rs
tbrezot: I have implemented a solution without, but I am not sure this is really a win. I can't get it to compile without collecting after calling `as_blocks()` since it refers to a temporary variable.

This doesn't compile, though it would be great:
```
let blocks = chain
    .into_iter()
    .flat_map(|(_, chain_value)| chain_value.as_blocks()); // <- cannot return reference to local variable `chain_value` returns a reference to data owned by the current function
reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks)?);
```

This compiles:
```
let blocks: Vec<Block<BLOCK_LENGTH>> = chain
    .into_iter()
    .flat_map(|(_, chain_value)| chain_value.into_blocks())
    .collect();
reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks.iter())?);
```

src/structs.rs
tbrezot: It is not possible to get a constant call to assert.

src/structs.rs
tbrezot: done in f2c5ec6

src/structs.rs
tbrezot: done in f2c5ec6

src/structs.rs
tbrezot: done in f2c5ec6

src/search.rs
tbrezot: same, cf f2c5ec6

src/chain_table.rs
tbrezot: Did something like that in f2c5ec6

src/structs.rs
ThibaudDauce: You can put "`BLOCK_LENGTH` should be *not* be greater than `MAX_BLOCK_LENGTH`"?

tbrezot: I think a user getting this error might want to know the numerical value rather than the name of the constant.

ThibaudDauce: As you wish, nobody will see this error anyway except developers :-D And `BLOCK_LENGTH` is already not shown in the error

ThibaudDauce (approved):

PR #34, opened 2023-03-30
Feature/live compact

tbrezot:
- implement live compacting (see trait documentation);
    - new compact trait;
    - allow having Entry Table lines associated with no Chain Table line (`last_chain_uid == None`);
- simplify upsert code

### DIFF
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -146,7 +146,7 @@ pub trait FindexCallbacks<Error: std::error::Error + CallbackError, const UID_LE
         new_chain_table_items: EncryptedTable<UID_LENGTH>,
     ) -> Result<(), Error>;
 
-    /// Returns all locations among the ones given that do not exist anymore.
+    /// Returns all locations that point to existing data.
     ///
     /// **NOTE**: this callback does not call the index database since indexed
     /// locations can be anything in Findex (DB UID, path, other ID...). It may
@@ -160,9 +160,14 @@ pub trait FindexCallbacks<Error: std::error::Error + CallbackError, const UID_LE
         locations: &HashSet<Location>,
     ) -> Result<HashSet<Location>, Error>;
 
+    #[cfg(feature = "live_compact")]
     /// Returns all locations that point to existing data.
     fn filter_removed_locations(
         &self,
         locations: &HashSet<Location>,
     ) -> Result<HashSet<Location>, Error>;
+
+    #[cfg(feature = "live_compact")]
+    /// Delete the Chain Table lines with the given UIDs.
+    async fn delete_chain(&mut self, uids: &HashSet<Uid<UID_LENGTH>>) -> Result<(), Error>;
 }
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -159,4 +159,10 @@ pub trait FindexCallbacks<Error: std::error::Error + CallbackError, const UID_LE
         &self,
         locations: &HashSet<Location>,
     ) -> Result<HashSet<Location>, Error>;
+
+    /// Returns all locations that point to existing data.
+    fn filter_removed_locations(
+        &self,
+        locations: &HashSet<Location>,
+    ) -> Result<HashSet<Location>, Error>;
 }
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -138,12 +138,11 @@ pub trait FindexCompact<
 
         // Get the values stored in the reindexed chains.
         let mut reindexed_chain_values = HashMap::with_capacity(chains_to_reindex.len());
-        for (kwi, chain) in chains_to_reindex {
+        for (kwi, chain) in &chains_to_reindex {
             let blocks = chain
-                .into_iter()
-                .flat_map(|(_, chain_value)| chain_value.into_blocks())
-                .collect::<Vec<_>>();
-            reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks.iter())?);
+                .iter()
+                .flat_map(|(_, chain_value)| chain_value.as_blocks());
+            reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks)?);
         }
 
         //
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -91,10 +91,8 @@ pub trait FindexCompact<
         // `chain_table_uids_to_remove`) and send them to the callback to update the
         // database.
         let mut chain_table_adds = EncryptedTable::default();
-        let mut entry_table = EntryTable::decrypt::<BLOCK_LENGTH, DEM_KEY_LENGTH, DemScheme>(
-            &k_value,
-            &encrypted_entry_table,
-        )?;
+        let mut entry_table =
+            EntryTable::decrypt::<DEM_KEY_LENGTH, DemScheme>(&k_value, &encrypted_entry_table)?;
         // Entry Table items indexing empty chains should be removed.
         let mut entry_table_uids_to_drop = Vec::new();
 
@@ -232,8 +230,7 @@ pub trait FindexCompact<
         // Call the callback
         self.update_lines(
             chain_table_uids_to_remove,
-            entry_table
-                .encrypt::<BLOCK_LENGTH, DEM_KEY_LENGTH, DemScheme>(&new_k_value, &mut rng)?,
+            entry_table.encrypt::<DEM_KEY_LENGTH, DemScheme>(&new_k_value, &mut rng)?,
             chain_table_adds,
         )?;
 
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -150,7 +150,7 @@ pub trait FindexCompact<
         // calling for one location batch for one word to add noise and prevent
         // the database the size of the chains for each keywords.
         //
-        let removed_locations = self.list_removed_locations(
+        let removed_locations = self.filter_removed_locations(
             &reindexed_chain_values
                 .values()
                 .flat_map(|chain| chain.iter().filter_map(IndexedValue::get_location))
--- a/src/compact.rs
+++ b/src/compact.rs
@@ -10,7 +10,7 @@ use cosmian_crypto_core::{
 use rand::seq::IteratorRandom;
 
 use crate::{
-    chain_table::{ChainTableValue, KwiChainUids},
+    chain_table::{ChainTable, ChainTableValue, KwiChainUids},
     entry_table::{EntryTable, EntryTableValue},
     error::CallbackError,
     parameters::check_parameter_constraints,
@@ -90,9 +90,10 @@ pub trait FindexCompact<
         // The goal of this function is to build these two data sets (along with
         // `chain_table_uids_to_remove`) and send them to the callback to update the
         // database.
-        let mut chain_table_adds = EncryptedTable::default();
         let mut entry_table =
             EntryTable::decrypt::<DEM_KEY_LENGTH, DemScheme>(&k_value, &encrypted_entry_table)?;
+        let mut chain_table_adds = EncryptedTable::default();
+
         // Entry Table items indexing empty chains should be removed.
         let mut entry_table_uids_to_drop = Vec::new();
 
@@ -197,8 +198,6 @@ pub trait FindexCompact<
             let mut new_entry_table_value = EntryTableValue::<UID_LENGTH, KWI_LENGTH>::new::<
                 CHAIN_TABLE_WIDTH,
                 BLOCK_LENGTH,
-                KMAC_KEY_LENGTH,
-                KmacKey,
             >(&mut rng, entry_table_value.keyword_hash);
 
             // Derive the new keys.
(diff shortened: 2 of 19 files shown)


src/callbacks.rs
```diff
@@ -146,7 +146,7 @@ pub trait FindexCallbacks<Error: std::error::Error + CallbackError, const UID_LE
         new_chain_table_items: EncryptedTable<UID_LENGTH>,
     ) -> Result<(), Error>;
 
-    /// Returns all locations among the ones given that do not exist anymore.
```
JosePisco, line 149: Question coding style: les commentaires avec trois '/' sont voulus ?

src/compact.rs
```diff
@@ -139,12 +138,11 @@ pub trait FindexCompact<
...
-                .into_iter()
-                .flat_map(|(_, chain_value)| chain_value.into_blocks())
-                .collect::<Vec<_>>();
-            reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks.iter())?);
+                .iter()
+                .flat_map(|(_, chain_value)| chain_value.as_blocks());
+            reindexed_chain_values.insert(kwi.clone(), IndexedValue::from_blocks(blocks)?);
         }
 
         //
```
JosePisco, line 148: Unused comment

src/compact_live.rs
```diff
@@ -0,0 +1,449 @@
...
+                "noise ration cannot be greater than 1 ({} given)",
+                Self::NOISE_RATIO
+            )));
+        }
+
+        let uids = self.fetch_all_entry_table_uids().await?;
+
+        // TODO (TBZ): find which probability is associated to this formula.
+        let entry_table_length = uids.len() as f64;
+        let n_compact = ((entry_table_length * (entry_table_length.log2() + 0.58))
```
JosePisco, line 125: Magic number - related to the TODO ?

src/compact_live.rs
```diff
@@ -0,0 +1,449 @@
...
+                "Number of Entry Table UIDs to compact ({n_compact}) should not be greater than {}",
+                uids.len() as f64 / (1f64 + Self::NOISE_RATIO)
+            )));
+        }
+
+        // Upper bounds are used for capacity in order to avoid reallocations:
+        // - the number of noise UIDs is lower than `n_compact`;
+        let mut noise = HashSet::with_capacity(n_compact);
+        // - the total number of UIDs (target set + noise set) is lower than `2*n_compact`.
+        let mut noisy_uids_to_compact = Vec::with_capacity(2 * n_compact );
```
JosePisco, line 141: Trailing whitespace within parenthesis

src/callbacks.rs
ThibaudDauce: Yes, `///` is the Rust way to comment function / trait / struct (docblocks) `//` are for inline comment inside code.

src/compact_live.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// - `num_reindexing_before_full_set`  : number of compact operations need to compact all the
+    /// base
+    async fn select_noisy_uids(
+        &self,
+        rng: &mut impl CryptoRngCore,
+        num_reindexing_before_full_set: u32
+    ) -> Result<(Vec<Uid<UID_LENGTH>>, HashSet<Uid<UID_LENGTH>>), Error<CustomError>> {
+        if Self::NOISE_RATIO > 1f64 {
+            return Err(Error::CryptoError(format!(
+                "noise ration cannot be greater than 1 ({} given)",
```
ThibaudDauce, line 116:
```suggestion
                "noise ratio cannot be greater than 1 ({} given)",
```

src/compact.rs
tbrezot: This is the "start" of a comment written by @ThibaudDauce:
```        //
        // Call `list_removed_locations` for all words in one pass instead of
        // calling for one location batch for one word to add noise and prevent
        // the database the size of the chains for each keywords.
        //
```

I do not especially appreciate this style of comment, but we do not have any defined style at Cosmian.

ThibaudDauce: Yes sometimes I like to add an empty line before a big comment but not attached to it if you want to remove that.

src/compact_live.rs
tbrezot: This is the formula used to compute the size of the samples to draw in order to have drawn all the set after `num_reindexing_before_full_set` samples. This formula contains "magic numbers" but you are right, I will link an explanation (this was partly was I meant when writing my todo).

src/compact_live.rs
tbrezot: Yes. I don't know why, it seems Rustfmt does not work anymore on this file on my machine...

src/entry_table.rs
```diff
@@ -179,21 +179,25 @@ impl<const UID_LENGTH: usize, const KWI_LENGTH: usize> EntryTableValue<UID_LENGT
     ///
     /// - `k_value`     : `𝐾_value`
```
JosePisco, line 180: Non unicode character '𝐾' in comment

src/entry_table.rs
```diff
@@ -245,7 +253,7 @@ impl<const UID_LENGTH: usize, const KWI_LENGTH: usize> EntryTableValue<UID_LENGT
             entry.push(current_chain_table_uid.clone());
 
             // return if we found the UID stored in the Entry Table value
```
JosePisco, line 255: Missing uppercase / period in comment.

src/entry_table.rs
tbrezot: Yes, I originally used a copy/paste from the LaTeX PDF, thus this fancy K. Is it a bad thing ?

src/entry_table.rs
```diff
@@ -227,12 +231,16 @@ impl<const UID_LENGTH: usize, const KWI_LENGTH: usize> EntryTableValue<UID_LENGT
...
             .entry(self.kwi.clone())
             .or_insert_with(Vec::new);
 
+        if self.chain_table_uid.is_none() {
+            return;
+        }
+
+        let kwi_uid: KmacKey = self.kwi.derive_kmac_key(CHAIN_TABLE_KEY_DERIVATION_INFO);
+
         // derive the Chain Table UID
```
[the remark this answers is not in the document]
tbrezot: Maybe you can suggest change for this kind of remark:
```suggestion
        // Derive the Chain Table UID.
```
Otherwise I may be too lazy to correct all comments one by one. Especially if this is not new code.

src/compact.rs
JosePisco: Did not see the following comments, github squished them

src/entry_table.rs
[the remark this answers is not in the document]
JosePisco: Great feature, will do.

src/entry_table.rs
JosePisco: Non-unicode characters may be undefined / different on other os / architectures.
This is a comment and I'm pretty sure rust compiler wouldn't mind it, but still, in a concern of portability I would avoid those as much as possible
```suggestion
    /// - `k_value`     : `K_value`
```

src/entry_table.rs
JosePisco:
```suggestion
            // Return if we found the UID stored in the Entry Table value.
```

src/compact_live.rs
JosePisco: I was thinking more like defining a constant with an explicit name to avoid useless cognitive complexity for the code.

tbrezot: Done in ee25dc7

src/compact_live.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// - `encrypted_entry_table`   : encrypted Entry Table
+    async fn fetch_chain_data(
+        &self,
+        k_value: &DemScheme::Key,
+        encrypted_entry_table: &EncryptedTable<UID_LENGTH>,
+    ) -> Result<
+        (
+            HashMap<Uid<UID_LENGTH>, Vec<Uid<UID_LENGTH>>>,
+            HashMap<Uid<UID_LENGTH>, HashSet<IndexedValue>>,
+        ),
```
ThibaudDauce, line 189: Could you create a small struct instead of a 4 elements tuple?

tbrezot: Done in 899799d

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+            master_key.derive_dem_key::<DEM_KEY_LENGTH, DemScheme::Key>(ENTRY_TABLE_KEY_DERIVATION_INFO);
+
+        let (noisy_entry_table_uids, noise) = self.select_noisy_uids(&mut rng, num_reindexing_before_full_set).await?;
+
+        // Compact the selected Entry Table UIDs by batch (noisy).
+        for noisy_uids_to_compact in noisy_entry_table_uids.chunks(Self::BATCH_SIZE) {
+            self.live_compact_uids(
+                &mut rng,
+                &k_value,
+                noisy_uids_to_compact.to_vec(),
```
ThibaudDauce, line 447: Can we take a chunk here instead of reallocating?

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+        &mut self,
+        master_key: &KeyingMaterial<MASTER_KEY_LENGTH>,
+        num_reindexing_before_full_set: u32,
+    ) -> Result<(), Error<CustomError>> {
+        let mut rng = CsRng::from_entropy();
+
+        let k_value =
+            master_key.derive_dem_key::<DEM_KEY_LENGTH, DemScheme::Key>(ENTRY_TABLE_KEY_DERIVATION_INFO);
+
+        let (noisy_entry_table_uids, noise) = self.select_noisy_uids(&mut rng, num_reindexing_before_full_set).await?;
```
ThibaudDauce, line 440: I think the `noisy_entry_table_uid` is not necessary, could be `entries_uids` because `noisy` is really similar than `noise` but means something totally different in this case (I prefer to not have it for a better comprehension)

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+        &mut self,
+        rng: &mut impl CryptoRngCore,
+        k_value: &DemScheme::Key,
+        noisy_uids: Vec<Uid<UID_LENGTH>>,
+        noise: &HashSet<Uid<UID_LENGTH>>,
+    ) -> Result<(), Error<CustomError>> {
+        // Fetch all index entries (noisy).
+        let mut noisy_encrypted_entry_table = self
+            .fetch_entry_table(&HashSet::from_iter(noisy_uids.into_iter()))
+            .await?;
```
ThibaudDauce, line 340: I think you should fetch all the entries here (required uids and the ones from the noise) because otherwise you leak which keywords will be updated.

ThibaudDauce: My bad, I didn't see that `noisy_uids` contains the noise one… Not sure about it, it's wasn't clear for me…

ThibaudDauce: I think it's better to have only the real uids to compact inside this Vec, and merge the two sets when calling `fetch_entry_table`

ThibaudDauce:
I think there is a lot of variables with `noisy_` and comments with `… (noisy)`, not sure if it gives much information to the reader (it was a little bit hard to read for me)

src/compact_live.rs
tbrezot: I thought it was clearer to add the `noisy_` prefix. This indicates that a variable which names is prefixed by it contains both targeted and noise data.

ThibaudDauce: It was not clear for me. I think it's better to have two separate variable (and it avoid duplicate memory too)

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+                    .iter()
+                    .filter_map(|(uid, chains)| {
+                        if noisy_encrypted_entry_table.contains_key(uid) {
+                            Some(chains.keys())
+                        } else {
+                            None
+                        }
+                    })
+                    .flatten()
+                    .cloned(),
```
ThibaudDauce, line 398: I think a for loop here will be clearer (and shorter)
```suggestion
            for (uid, chains) in new_chains {
                if noisy_encrypted_entry_table.contains_key(&uid) {
                    chains_to_delete.extend(chains.0.into_keys());
                }
            }
```

It requires to `pub` the HashMap inside `EncryptedTable` (to access `.0`) or to have a implementation to take ownership of the inner HashMap (we currently have only function to get a reference to the inner HashMap). It also avoid the `.cloned()` at the end. No allocation needed since the `new_chains` is only used once

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+                    .filter_map(|(uid, chain)| {
+                        if !noisy_encrypted_entry_table.contains_key(uid) && !noise.contains(uid) {
+                            Some(chain)
+                        } else {
+                            None
+                        }
+                    })
+                    .flatten()
+                    .cloned(),
+            );
```
ThibaudDauce, line 414: We may be able to do the same here (the for loop should be smaller) but the `.clone()` is required since `chains` is re-used multiple times inside the while loop.

src/compact_live.rs
ThibaudDauce: I prefer something like `noise_and_target_uids` / `noise_uids` / `target_uids` in that case. `noisy` could be easily interpreted as "only noise" or "in random order" or "fetched noisly" like in `noisy_fetch_chains` in search operation.

src/compact_live.rs
```diff
@@ -0,0 +1,454 @@
...
+        // Associate Entry Table UIDs to Chain Table UIDs.
+        for (uid, v) in  entry_table.iter() {
+            chains.chain_uids.insert(
+                uid.clone(),
+                kwi_chain_table_uids
+                .get(&v.kwi)
+                .ok_or(Error::<CustomError>::CryptoError(format!(
+                    "no matching Kwi in `kwi_chain_table_uids` ({:?})",
+                    v.kwi
+                )))?.clone(),
```
ThibaudDauce, line 213: Since `kwi_chain_table_uids` is only used once (and I think `v.kwi` is unique?) you may be able to call `.remove()` instead of `.get` which take ownership of the value and avoid the `.clone()` at the end.

ThibaudDauce: You need to call 
```
        let chain_values = self
            .batch_fetch_chains(&kwi_chain_table_uids)
            .await?
```
before by doing it since the `kwi_chain_table_uids` will be empty after the first loop…

src/compact_live.rs
tbrezot: I changed to `mixt_uids` and `noise_uids` and described in several comments the meaning of `mixt_`.

src/compact_live.rs
tbrezot: Good point.

We need to allocate at some point to pass a `HashSet` to the fetch. I chose to change the signature to take this hashset as argument. This avoids one allocation.

src/upsert.rs
```diff
@@ -58,64 +60,59 @@ pub trait FindexUpsert<
...
+                new_chains
+                    .entry(keyword.hash())
                     .or_default()
                     .insert(indexed_value.clone(), BlockType::Addition);
             }
         }
-
         // Adding and deleting the same indexed value for the same keyword only performs
-        // the deletion.
+        // the deletion. Adding deletions after additions only keeps deletions.
```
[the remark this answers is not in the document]
tbrezot: Right.

src/compact_live.rs
tbrezot: Changed to `mixt_uids`. The merging of the target and noise sets can be implemented in the `for` loop in a later time.

src/compact_live.rs
tbrezot: done in 9d26714

src/compact_live.rs
tbrezot: done in 021cf5b

ThibaudDauce (approved):

PR #51, opened 2023-05-02
fix: remove the `max_uid_per_chain` parameter

tbrezot:


src/entry_table.rs
```diff
@@ -224,11 +226,10 @@ impl<const UID_LENGTH: usize, const KWI_LENGTH: usize> EntryTableValue<UID_LENGT
     >(
         &self,
         kwi_uid: &KmacKey,
-        max_uid_per_chain: usize,
     ) -> Vec<Uid<UID_LENGTH>> {
         let mut chain = Vec::new();
 
-        while chain.len() < max_uid_per_chain && chain.last() != self.chain_table_uid.as_ref() {
+        while chain.len() < MAX_UID_PER_CHAIN && chain.last() != self.chain_table_uid.as_ref() {
```
ThibaudDauce, line 232: Should we fail when reaching `MAX_UID_PER_CHAIN`? Here we'll silently return incomplete results, no?

tbrezot: indeed, see 7950df0

ThibaudDauce (approved):

src/entry_table.rs
```diff
@@ -23,6 +23,9 @@ use crate::{
     KeyingMaterial, Keyword, CHAIN_TABLE_KEY_DERIVATION_INFO,
 };
 
+/// Maximum number of Chain Table UID associated to a given Entry Table UID.
+const MAX_UID_PER_CHAIN: usize = 1_000_000;
```
Manuthor, line 27: I suppose 1_000_000 is a reasonable limit. What happens if this limit is reached at one point? no way to override it right?

ThibaudDauce: I think Findex will be unusable way before this limit (timing / size wise). Users shouldn't index so much locations inside the same keyword.

PR #53, opened 2023-05-09
Feature/wrap upsert

tbrezot:


src/upsert.rs
```diff
@@ -34,59 +32,81 @@ pub trait FindexUpsert<
...
+    ///
+    /// - `master_key`      : Findex master key
+    /// - `label`           : additional public information used in key hashing
+    /// - `modifications`   : set of operations to apply on the associated value
+    ///   for the associated keyword
+    async fn upsert(
+        &mut self,
+        master_key: &KeyingMaterial<MASTER_KEY_LENGTH>,
+        label: &Label,
+        modifications: HashMap<Keyword, HashMap<IndexedValue, BlockType>>,
```
ThibaudDauce, line 97: I'm not sure I like the exposition of `BlockType` to the client. I may prefer the old way with two parameters… What do you think @Manuthor?

Manuthor: I suppose it has been made to avoid an allocation. IMHO, additions and deletions are more readable

tbrezot: I was thinking of it as a way to remove this ugly piece of code from the `upsert`:
```        let mut new_chains =
            HashMap::<KeywordHash, HashMap<IndexedValue, BlockType>>::with_capacity(
                additions.len() + deletions.len(),
            );
        for (indexed_value, keywords) in additions {
            for keyword in keywords {
                new_chains
                    .entry(keyword.hash())
                    .or_default()
                    .insert(indexed_value.clone(), BlockType::Addition);
            }
        }
        for (indexed_value, keywords) in deletions {
            for keyword in keywords {
                new_chains
                    .entry(keyword.hash())
                    .or_default()
                    .insert(indexed_value.clone(), BlockType::Deletion);
            }
        }
```

tbrezot: But I removed in it the last commits.

ThibaudDauce (approved):

PR #54, opened 2023-05-12
Allow multiple chains for one keyword

ThibaudDauce:
- [ ] fix `contains_uid` and `get` method not optimized for `Vec`
- [ ] is it better to allow only entries to be duplicated and not chains?

### DIFF
--- a/src/core/structs.rs
+++ b/src/core/structs.rs
@@ -365,16 +365,41 @@ impl_byte_array!(Uid);
 // NOTE TBZ: need struct to implement `Serializable`
 #[must_use]
 #[derive(Default, Debug, Clone)]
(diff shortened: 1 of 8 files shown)


src/core/structs.rs
```diff
@@ -365,16 +365,36 @@ impl_byte_array!(Uid);
 // NOTE TBZ: need struct to implement `Serializable`
 #[must_use]
 #[derive(Default, Debug, Clone)]
-pub struct EncryptedTable<const UID_LENGTH: usize>(HashMap<Uid<UID_LENGTH>, Vec<u8>>);
+pub struct EncryptedTable<const UID_LENGTH: usize>(Vec<(Uid<UID_LENGTH>, Vec<u8>)>);
```
Manuthor, line 368: we could declare an `EncryptedTableLine` struct for the type `(Uid<UID_LENGTH>, Vec<u8>)`

PR #57, opened 2023-05-15
Feature/refactoring

tbrezot:
A branch to implement Findex's new look following the work on the formalization with @chloehebant.

In order to ease the reading, let me fix some vocabulary first:
- Encrypted Dictionary (EDX): a key value store which values are of constant size;
- Encrypted Multi-Map (EMM): a key value store which values are of variable size;
- Encrypted Graph (EGX): an encrypted graph which nodes contain data and pointers to other nodes;
- Encrypted Dictionary Scheme (DX-Enc): a scheme managing an EDX;
- Encrypted Multi-Map Scheme (MM-Enc): a scheme managing an EMM;
- Encrypted Graph Scheme (GX-Enc): a scheme managing an EGX;
- tag: bytes (may be a meaningful piece of information) used to point to a value in an map (H(w) in the Entry Table)
- token: *non-meaningul* bytes used to index a value in a map (it corresponds to the UIDs)

Findex (as the product) is now composed of three algorithms:
- Findex: an index interface to hide the cryptographic details of Findex Graph;
- Findex Graph: a GX-Enc scheme using a MM-Enc scheme;
- Findex Multi-Map: a MM-Enc scheme using two DX-Enc schemes.

Two generic DX-Enc schemes are used by Findex Multi-Map:
- Entry Table: an EDX scheme in charge of storing metadata about the chains: counter, key seed (Kwi), corresponding tag (H(w_i));
- Chain Table: an EMM used to store the actual chain data; it's implementation is actually done using an EDX.

**TODO**:
- [x] decide which constant to fix and which constant to use as generic (e.g. `BLOCK_LENGTH`/`LINE_LENGTH`)
- [x] add compact operation
- [x] pass all old tests
- [x] pass regression tests on database:
    - [x] add regression tests inside the repo (serialize in-memory database) 
    - [x] change counter back to the hash chain
- [x] make findex types `Sync + Send`
- [x] add encryption scheme to manage encryption
- [x] update CryptoCore version
- [x] add `progress` callback (injection)
- [x] add CATS compatibility


refactored/src/edx.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// For each new values:
+    /// - if there is no value in both the set of old values and the encrypted
+    ///   dictionary, inserts the new value;
+    /// - if there is a value in the set of old values that matches the
+    ///   value in the encrypted dictionary, replaces it by the new value;
+    /// - if there is a value in the set of old values that does not match the
+    ///   value in the encrypted dictionary, returns the encrypted value stored;
+    /// - if there is a value in the set of old values but no value in the
+    ///   encrypted dictionary, returns an error.
```
ThibaudDauce, line 57: Maybe clearer with a markdown table?

| old\existing | None | Some("A")  | Some("B")  |
|--------------|------|------------|------------|
| None         |   ?  |    Error   |    Error   |
| Some("A")    |   ?  | Insert new |   Reject   |
| Some("B")    |   ?  |   Reject   | Insert new |

refactored/src/edx.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// All modifications to the encrypted dictionary are *atomic*.
```
ThibaudDauce, line 59: Maybe explain that it's not the complete `upsert` that is atomic (if we reject some values because the old values don't match we can still insert other values from the same `upsert`)

refactored/src/edx.rs
```diff
@@ -0,0 +1,75 @@
...
+        &self,
+        k: &Self::Key,
+        old_values: &HashMap<Self::Token, Self::EncryptedValue>,
+        new_values: HashMap<Self::Token, Self::Value>,
+    ) -> Result<HashMap<Self::Token, Self::EncryptedValue>, Error>;
+
+    /// Encrypts the given values and inserts them into the Encrypted Dictionary
+    /// if no value is already stored for the corresponding tokens. Returns the
+    /// set of tokens for which a value is already stored.
+    fn insert(
```
ThibaudDauce, line 70: Maybe rename this function `insert_missing`? Since it allows only filling empty slots?

refactored/src/emm.rs
```diff
@@ -0,0 +1,39 @@
...
+use zeroize::ZeroizeOnDrop;
+
+pub trait Emm<const KEY_LENGTH: usize, const TOKEN_LENGTH: usize> {
+    type Key: ZeroizeOnDrop;
+
+    /// Cryptographically secure token used to index values inside the encrypted
+    /// multi-map.
+    type Token = [u8; TOKEN_LENGTH];
+
+    /// Fixed length value stored inside the dictionary.
```
ThibaudDauce, line 13: Not fixed length :-)

refactored/src/emm.rs
```diff
@@ -0,0 +1,39 @@
...
+    ///
+    /// One network call should be made for all the tokens.
+    fn get(
+        k: &Key<KEY_LENGTH>,
+        tokens: HashSet<Self::Token>,
+    ) -> HashMap<Self::Token, Vec<Self::Value>>;
+
+    /// Encrypts the given values using the given key and insert the ciphertexts
+    /// into the multi-map.
+    fn insert(k: &Key<KEY_LENGTH>, values: HashSet<Self::Token, Vec<Self::Value>>);
```
ThibaudDauce, line 38: `HashMap`, no?
```suggestion
    fn insert(k: &Key<KEY_LENGTH>, values: HashMap<Self::Token, Vec<Self::Value>>);
```

refactored/src/entry_table.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn encrypt(
+        key: &[u8],
+        value: [u8; VALUE_LENGTH],
+    ) -> Result<Self, cosmian_crypto_core::CryptoCoreError> {
+        let mut res = Self {
+            ciphertext: value,
+            nonce: [0; NONCE_LENGTH],
+            tag: [0; MAC_LENGTH],
+        };
+        let mut rng = CsRng::from_entropy();
```
ThibaudDauce, line 40: Maybe accept a `CsRng` in trait parameters to avoid creating new one during each encrypt?

refactored/src/edx.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Deterministically transforms the given tag into a token using the given
+    /// key. The token should leak no information about the tag.
+    fn tokenize(k: &Self::Key, tag: &[u8]) -> Self::Token;
+
+    /// Queries the encrypted dictionary for the given tokens and returns the
+    /// encrypted values.
+    ///
+    /// One network call should be made for all the tokens in order to improve
+    /// the performances.
```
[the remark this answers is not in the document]
tbrezot: removed

refactored/src/edx.rs
tbrezot: Actually the entire `upsert` should be atomic if the following property I've added is respected:
```
    /// # Error
    ///
    /// An error should be return when a value is being upserted for a token for
    /// which the EDX contains no value but an old value is passed. If this
    /// happens, no value should be upserted at all.
```

It is unclear to me what is the desired behavior in this case. This is not supposed to be a possible case, hence the returned error.

refactored/src/emm.rs
tbrezot: Changed

refactored/src/emm.rs
tbrezot: Yes, this file was temporary not declared in the `lib.rs`, hence the erroneous type...

refactored/src/entry_table.rs
tbrezot: I moved the CsRng into the `Index` struct to avoid generating a new one in each call to `add` or `delete` (which the old version was doing).

refactored/src/edx.rs
tbrezot: Good point :+1:

refactored/src/edx.rs
ThibaudDauce: But if an `old_value` doesn't match what is inside the database, other values (where `old_value` match) will be updated?

tbrezot: We may not want to update a corrupted index?

refactored/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+//! the interface and to hide the cryptographic details of the implementation
+//! when it is possible.
+//!
+//! However, since Findex is database agnostic and is generic on the choice of
+//! the Dx-Enc algorithm, some generic parameters are still exposed through the
+//! `Index` interface.
+
+#![feature(never_type)]
+#![allow(incomplete_features)]
+#![feature(generic_const_exprs)]
```
Manuthor, line 18: as discussed, is this feature finally required?

tbrezot: removed in 9712ff0

refactored/src/findex_mm/structs.rs
```diff
@@ -0,0 +1,247 @@
...
+        SEED_LENGTH,
+        KEY_LENGTH,
+        TOKEN_LENGTH,
+        { 1 + (BLOCK_LENGTH + 1) * LINE_LENGTH },
+        ChainTableError,
+    >,
+> From<EntryTableValue<BLOCK_LENGTH, LINE_LENGTH, ChainTableError, ChainTable>>
+    for [u8; ENTRY_TABLE_VALUE_LENGTH]
+where
+    [(); 1 + (BLOCK_LENGTH + 1) * LINE_LENGTH]:,
```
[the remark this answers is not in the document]
tbrezot: removed in 9712ff0

refactored/src/edx/entry_table.rs
```diff
@@ -0,0 +1,178 @@
...
+    type Token = [u8; TOKEN_LENGTH];
+
+    fn setup() -> Self {
+        Self {
+            edx: Edx::setup(),
+            err: PhantomData,
+        }
+    }
+
+    fn gen_seed(&self, rng: &mut impl CryptoRngCore) -> Self::Seed {
```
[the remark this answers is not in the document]
tbrezot: It is currently not possible as default type are not allowed in traits. I would need a trait defining a method `new()` for  `Self::Seed` and `from_seed()` for `Self::Key`.

Since the amount of duplicated code is not that big, I prefer to let it as is.

refactored/src/edx.rs
tbrezot: Done in 76b02e7

refactored/src/edx.rs
tbrezot: This is actually a much more difficult problem than I thought (cf [this document](https://www.notion.so/cosmian/Findex-regeneration-and-integrity-properties-fb4371c5171a4c949a2c59675347e8c6)). This won't be solve for this PR. I propose to raise an issue and solve this on a later time.

src/edx/chain_table.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    fn setup(edx: Self::Store) -> Self {
+        Self(edx)
+    }
+
+    fn gen_seed(&self, rng: &mut impl CryptoRngCore) -> Self::Seed {
+        Seed::new(rng)
+    }
+
+    fn derive_keys(&self, seed: &Self::Seed) -> Self::Key {
```
JosePisco, line 60: When deriving keys, why do you pass constants in parameters  ? `CHAIN_TABLE_KEY_DERIVATION_INFO` and `b"KMAC key"` ? Why do you need to update the hash with public constants ?

src/edx/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// Fixed length encrypted value stored inside the encrypted dictionary.
+    type EncryptedValue: Debug + Sized + Send + Sync + Clone;
+
+    type Store: EdxStore<VALUE_LENGTH>;
+
+    /// Instantiates a new Dx-Enc scheme.
+    fn setup(edx: Self::Store) -> Self;
+
+    /// Generates a new random seed.
+    fn gen_seed(&self, rng: &mut impl CryptoRngCore) -> Self::Seed;
```
JosePisco, line 60: Is the seed supposed to be cryptographically secure, i.e. generated by a CSPRNG ?

src/findex_graph/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        Value: Hash + Eq + Clone + From<Vec<u8>>,
+    >(
+        _res: HashMap<Tag, HashSet<IndexedValue<Tag, Value>>>,
+    ) -> bool {
+        false
+    }
+
+    #[actix_rt::test]
+    async fn test_insert_get() {
+        let rng = Arc::new(Mutex::new(CsRng::from_entropy()));
```
JosePisco, line 126: I think I recall a discussion we had on chacha 12 rounds for csprng purposes, however I can't remember if we agreed on it being enough rounds.

JosePisco: With source https://eprint.iacr.org/2019/1492 saying an attack against chacha7 takes roughly 2^238 time complexity, we can safely say that 12 rounds is a great deal between efficiency and security.

We should still note that it doesn't follow any standard.

src/edx/chain_table.rs
tbrezot: These constants are used as domain separation information. Different constants are used for the Entry Table.

src/edx/mod.rs
tbrezot: It is since it is used to generate the keys. The RNG passed here implements `CryptoRngCore` which makes it a CSPRNG.

src/edx/chain_table.rs
JosePisco: Oh yes, because you derive two times from the same initial key. Didn't see, makes sense :+1:

src/index/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+            return tokens.iter().cloned().collect();
+        }
+
+        let mut rng = self.rng.lock().expect("could not lock mutex");
+        let mut res = HashSet::with_capacity(n);
+        for _ in 0..n {
+            // In order to draw a random element from the set, draw a random u64 and use it
+            // modulo the length of the set. This is not perfectly uniform but should be
+            // enough in practice.
+            let index = (rng.next_u64() % tokens.len() as u64) as usize;
```
JosePisco, line 230: iirc, RNG exposes `gen_range(a..b)` method to draw from a uniform distribution.

JosePisco (dismissed):

src/index/mod.rs
tbrezot: I do not want to depend on the `Rng` trait (which would require compiling the whole `rand` module), but `RngCore` which does not implement `gen_range` (but only requires compiling `rand_core` module).

Manuthor (approved):

PR #61, opened 2023-06-02
fix: race condition in `fetch_chains()`

tbrezot:
If an upsert modifies the ET and a search is performed before the new values are inserted in the CT, the search client derives UIDs that do not exist yet inside the CT and tries fetching them. When trying to recompose the chain, the removed check would raise an error.

Findex only implements eventual consistency between the two tables, thus no error should not be raised.

### DIFF
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -219,30 +219,27 @@ pub trait FetchChains<
             let mut chain = Vec::with_capacity(chain_table_uids.len());
 
             for uid in chain_table_uids {
-                let encrypted_value =
-                    encrypted_items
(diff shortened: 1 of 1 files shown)


Manuthor (approved):

ThibaudDauce:
Hum… Maybe we could raise an error if it's not the last element missing?

tbrezot:
> Hum… Maybe we could raise an error if it's not the last element missing?

I thought about it, but this is not enough since another upsert client could have added other values *after* the missing ones.

ThibaudDauce (approved):

PR #65, opened 2023-06-23
test: impl Display trait on structs used in callbacks for logging

Manuthor:


src/structs.rs
```diff
@@ -435,14 +436,64 @@ impl Serializable for IndexedValue {
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl<const UID_LENGTH: usize> Display for Uids<UID_LENGTH> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut output = String::new();
+        for uid in self.clone() {
+            output = format!("uid: {}\n{output}", STANDARD.encode(uid));
```
tbrezot, line 471: Is this really what you want ? I mean it seems to me that only the last UID is displayed, but maybe there is something I didn't get.

Manuthor: it also takes the `output` variable for each element of the list

tbrezot: Ah okay, I got it, you include `output` in the `format!()`... why don't you simply extend the output with the new `format` ?

Manuthor: done in https://github.com/Cosmian/findex/pull/65/commits/9b7dde9771647d197ad4474ccdec55e0be60bd8a

PR #68, opened 2023-07-21
Fix/make compact callback async

tbrezot:
fix: make compact callbacks async
    
    - `update_lines`
    - `list_removed_locations`
    - `filter_removed_locations`
    
    Fix doc mistake for `list_removed_locations` (negation was missing)


bgrieder (dismissed):
LGTM
Should we remove the `#[cfg(feature = "live_compact")]` methods right now ?

tbrezot:
> LGTM Should we remove the `#[cfg(feature = "live_compact")]` methods right now ?

Done in 7049a3e

src/upsert.rs
```diff
@@ -34,12 +32,15 @@ pub trait FindexUpsert<
...
+    ///
+    /// Returns a map of keywords to booleans indicating whether the keyword
+    /// was already present in the database.
     async fn add(
         &mut self,
         master_key: &KeyingMaterial<MASTER_KEY_LENGTH>,
         label: &Label,
         items: HashMap<IndexedValue, HashSet<Keyword>>,
-    ) -> Result<(), Error<CustomError>> {
+    ) -> Result<HashMap<Keyword, bool>, Error<CustomError>> {
```
tbrezot, line 43: You just need to return the set of keyword that were not part of the index before the `add` operation. This is lighter and you can get your boolean using `.contains()`.

.DS_Store
tbrezot, line 1: Does this need to be removed?

src/upsert.rs
```diff
@@ -68,60 +72,79 @@ pub trait FindexUpsert<
...
-                    (keyword_hash, indexed_values),
-                )
+            .map(|(keyword, indexed_values)| {
+                let keyword_hash = keyword.hash();
+                let uid = EntryTable::<UID_LENGTH, KWI_LENGTH>::generate_uid::<KMAC_KEY_LENGTH>(
+                    &k_uid,
+                    &keyword_hash,
+                    label,
+                );
+                uid_to_keyword.insert(uid, keyword);
```
tbrezot, line 123: Do not add side effects to the maps. You could implement this using `unzip()` if you want to keep the functional style. Otherwise just use a `for` loop.

src/upsert.rs
```diff
@@ -68,60 +72,79 @@ pub trait FindexUpsert<
...
         let mut rng = CsRng::from_entropy();
         let k_uid = master_key.derive_kmac_key::<KMAC_KEY_LENGTH>(ENTRY_TABLE_KEY_DERIVATION_INFO);
         let k_value = master_key.derive_dem_key(ENTRY_TABLE_KEY_DERIVATION_INFO);
 
-        let mut new_chains =
-            HashMap::<KeywordHash, HashMap<IndexedValue, BlockType>>::with_capacity(
-                additions.len() + deletions.len(),
-            );
+        // a reverse map of Keyword UIDs (e.g. hashes) to Keywords
+        let mut uid_to_keyword = HashMap::<Uid<UID_LENGTH>, Keyword>::new();
```
tbrezot, line 92: You can reserve the adequate space here.
```suggestion
        let mut uid_to_keyword = HashMap::<Uid<UID_LENGTH>, Keyword>::with_capacity(additions.len() + deletions.len());
```

src/upsert.rs
```diff
@@ -68,60 +72,79 @@ pub trait FindexUpsert<
...
 
         let mut rng = CsRng::from_entropy();
         let k_uid = master_key.derive_kmac_key::<KMAC_KEY_LENGTH>(ENTRY_TABLE_KEY_DERIVATION_INFO);
         let k_value = master_key.derive_dem_key(ENTRY_TABLE_KEY_DERIVATION_INFO);
 
-        let mut new_chains =
-            HashMap::<KeywordHash, HashMap<IndexedValue, BlockType>>::with_capacity(
-                additions.len() + deletions.len(),
-            );
+        // a reverse map of Keyword UIDs (e.g. hashes) to Keywords
```
tbrezot, line 91: UIDs and hashes are not similar: an UID is a KMAC of a keyword. I think the variable name is clear enough to allow removing this comment anyway.

src/upsert.rs
```diff
@@ -68,60 +72,79 @@ pub trait FindexUpsert<
...
+        for uid in encrypted_entry_table.keys() {
+            if let Some(keyword) = uid_to_keyword.remove(uid) {
+                keyword_presence.insert(keyword, true);
+            }
+        }
+        // whatever is left in `uid_to_keyword` is not in the database,
+        // i.e. not already present. Update the keyword_presence accordingly
+        for (_, keyword) in uid_to_keyword {
+            keyword_presence.insert(keyword, false);
+        }
```
tbrezot, line 146: This is not useful since you can remove a set instead of a map keyword <-> boolean.

src/upsert.rs
```diff
@@ -68,60 +72,79 @@ pub trait FindexUpsert<
...
             .try_into()?;
 
+        // compute the map of keywords to booleans indicating whether the keyword
+        // was already present in the database
+        let mut keyword_presence = HashMap::new();
+        for uid in encrypted_entry_table.keys() {
+            if let Some(keyword) = uid_to_keyword.remove(uid) {
+                keyword_presence.insert(keyword, true);
+            }
+        }
```
tbrezot, line 141:
```suggestion
        let new_keywords = uid_to_keyword
            .into_iter()
            .filter(|(uid, _)| !encrypted_entry_table.contains_key(&uid))
            .map(|(_, keyword)| keyword)
            .collect::<HashSet<_>>();
```

tbrezot (commented):
Wouldn't it be better to track this commit using a dedicated branch to allow merging `fix/make compact callback async` into `develop`?

src/upsert.rs
bgrieder: let us go for a for loop; it is going to be hard to read with an unzip

french_dataset.json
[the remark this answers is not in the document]
tbrezot: I removed it.

.DS_Store
tbrezot: I removed it.

ackRow (approved):

PR #76, opened 2023-09-18
Fix/relax async req

tbrezot:
Using `#[async_trait]` enforce trait member outputs to implement `Send`, which is not possible when working with WASM in CloudproofRust.

Using `#[async_trait(?Send)]` allows relaxing this constraint.


Manuthor (dismissed):
Thanks for this

PR #86, opened 2024-07-15
fix: small rust review

Manuthor:


src/lib.rs
```diff
@@ -30,9 +30,6 @@ pub use kv::KvStore;
 /// mitigated by marking n bit of the addresses, allowing to statistically store up to 2^((64-n)/2)
 /// keywords, and reducing the number of words that can be used to store associated values to
 /// sqrt(2^64 - 2^n).
-#[cfg(not(feature = "small"))]
-pub const ADDRESS_LENGTH: usize = 16;
-#[cfg(feature = "small")]
```
tbrezot, line 35: I would like to keep this to work on it in the future

Manuthor: no problem, has been reverted.

PR #90, opened 2024-10-28
Feat/v7 redis

HatemMn:
### Running redis DB : 
`redis-server --port 9999`
9999 is the default port, choice was arbitrary

### Running the tests : 
`cargo test -- --test-threads=1` otherwise will fail


.vscode/extensions.json
tbrezot, line 1: not here

.vscode/settings.json
tbrezot, line 1: not here

benches/benches.rs
tbrezot, line 1: There should be no style-related diff, your editor is not properly configured.

src/address.rs
```diff
@@ -61,6 +66,21 @@ impl<const LENGTH: usize> Add<u64> for Address<LENGTH> {
...
+        write!(f, "{:?}", &self.0)
+    }
+}
+
+impl<const N: usize> ToRedisArgs for Address<N> {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        out.write_arg(self.to_string().as_bytes());
```
tbrezot, line 80: Very inefficient conversion.

src/findex.rs
```diff
@@ -265,4 +261,83 @@ mod tests {
             res
         );
     }
+
+    #[test]
+    fn test_insert_search_delete_search_redis() {
```
tbrezot, line 266: no redis-specific test in the main file : either make it backend-independent or move it to some redis-related module.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+        let safe_connection = &mut *self.connection.lock().expect("Poisoned lock.");
+        redis::cmd("FLUSHDB").exec(safe_connection)?;
+        Ok(())
+    }
+}
+
+/**
+ * The RedisMemory implementation of the MemoryADT trait.
+ * All operations are - and should be - atomic.
+ */
```
tbrezot, line 84: Use `///` to be consistent with the style of the repo.
use `--` instead of `-` for emphasis. Use back-quotes around variable/type/trait names

Otherwise, this doc carries no information.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+                },
+                Err(e) => panic!("Error creating redis client: {:?}", e),
+            },
+            _marker: PhantomData,
+        }
+    }
+}
+
+/**
+ * Flushes the Redis database.
```
tbrezot, line 69: This should be a function-level doc, not an implementation-level doc.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+ * WARNING: This is irreversible, do not run in production.
```
tbrezot, line 70: This is already ignored at compile-time for production-targeted binaries.
```suggestion
```

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+ */
+#[cfg(test)]
+impl<Address: Hash + Eq + Debug, const WORD_LENGTH: usize> RedisMemory<Address, WORD_LENGTH> {
+    pub fn flush_db(&self) -> Result<(), redis::RedisError> {
+        let safe_connection = &mut *self.connection.lock().expect("Poisoned lock.");
```
tbrezot, line 75: Do not store the reference.
```suggestion
        let mut cnx = self.connection.lock().expect("Poisoned lock.");
```

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Atomically reads the values at the given addresses.
+     */
+    async fn batch_read(
+        &self,
+        addresses: Vec<Address>,
+    ) -> Result<Vec<Option<Self::Word>>, Self::Error> {
+        let safe_connection = &mut *self.connection.lock().expect("Poisoned lock.");
+        let refs: Vec<&Address> = addresses.iter().collect(); // Redis MGET requires references to the values
```
tbrezot, line 100: Needless collect imho

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+        let res: Vec<Option<Self::Word>> =
```
tbrezot, line 101: Let types be inferred.
```suggestion
        let res =
```

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+            safe_connection.mget(&refs).map_err(|e| RedisMemoryError {
+                details: e.to_string(),
+            })?;
+        Ok(res)
```
tbrezot, line 105: Needless de-wrapping, wrapping of the error.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+     * Returns the current value on the guard's address.
+     * If the result is equal to the guard's old value, the bindings get
+     * written.
+     *
+     * Args that are passed to the LUA script are, in order :
+     * 1. Guard address.
+     * 2. Guard value.
+     * 3. Vector length.
+     * 4+. Vector elements (address, word).
+     */
```
tbrezot, line 119: This function is already documented in the trait. Plus your doc is wrong as you do not return "the current value of the guard's address".

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+    async fn guarded_write(
+        &self,
+        guard: (Self::Address, Option<Self::Word>),
+        bindings: Vec<(Self::Address, Self::Word)>,
+    ) -> Result<Option<Self::Word>, Self::Error> {
+        let safe_connection = &mut *self.connection.lock().expect("Poisoned lock.");
+        let (guard_address, guard_value) = guard;
+
+        const GUARDED_WRITE_LUA_SCRIPT: &str = r#"
```
tbrezot, line 128: Write this constant somewhere else.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        let script = redis::Script::new(GUARDED_WRITE_LUA_SCRIPT);
+        let mut script_invocation = script.prepare_invoke();
+
+        script_invocation.arg(guard_address);
+        if let Some(byte_array) = guard_value {
+            script_invocation.arg(&byte_array);
+        } else {
+            script_invocation.arg("false".to_string());
+        }
```
tbrezot, line 154: Not idiomatic. Use `unwrap_or` or something similar.

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+        script_invocation.arg(bindings.len());
+        for (address, word) in bindings {
+            script_invocation.arg(address).arg(&word);
+        }
+
+        let result: Result<Option<Self::Word>, redis::RedisError> =
+            script_invocation.invoke(safe_connection);
+
+        result.map_err(Into::into)
```
tbrezot, line 163: This should be a one-liner.
- do not explicit the type
- use `.into()` (may not even be needed)

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+        );
+
+        assert_eq!(
+            vec![Some([2])],
+            block_on(memory.batch_read(vec![1])).unwrap(),
+        );
+        memory.flush_db().unwrap(); // flush !
+
+        assert_eq!(vec![None], block_on(memory.batch_read(vec![1])).unwrap(),);
+        memory.flush_db().unwrap(); // prevent future tests from failing
```
tbrezot, line 186: no, each test should manage its own setup
```suggestion
```

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+            script_invocation.invoke(safe_connection);
+
+        result.map_err(Into::into)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_flush_db() {
```
tbrezot, line 170: why is this test before imports ?

tbrezot (commented):
I do not like having to mentally filter out unnecessary cosmetic changes. Fix your editor and review your own diffs to make sure they deserve to be read.

Many style comments, a few potential performance improvements. The core functionality seems correct. But this should be a one-file MR.

src/redis_memory.rs
HatemMn: is this better ?
```
        let res: Vec<Option<Self::Word>> = safe_connection
            .mget(addresses.as_slice())
            .map_err(RedisMemoryError::from)?;
        Ok(res)
```

src/redis_memory.rs
HatemMn: I kept this part because the new lua script is not so evident to read, is it ok ?
**/// Args that are passed to the LUA script are, in order:
/// 1. Guard address.
/// 2. Guard value. 
/// 3. Vector length.
/// 4+. Vector elements (address, word).**

and deleted the rest

src/redis_memory.rs
HatemMn: # UPDATE

Actually, the new syntax i used in the previous comment fails the tests so it's not valid

Do you have a valid alternative ?

src/redis_memory.rs
HatemMn:
```rust
        script_invocation.arg(&guard_value.unwrap_or("false".to_string()));
```

I beleive this is the idiomatic way of doing it, however it fails compilation because rust expects the same type for the "unwrap_or". I didn't find a shorter way

src/redis_memory.rs
HatemMn: New function
```
        script_invocation.invoke(&mut safe_connection).map_err(|e| e.into())
```

src/redis_memory.rs
HatemMn: fixed

src/redis_memory.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[ignore]
+    fn test_batch_read_error_handling() {
+        todo!("Not implemented");
+    }
+
+    #[test]
+    #[ignore]
+    fn test_guarded_write_error_handling() {
+        todo!("Not implemented");
+    }
```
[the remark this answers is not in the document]
HatemMn: ok

src/redis_memory.rs
tbrezot: ou ```
        script_invocation.invoke(&mut safe_connection).map_err(Into::into)```

src/redis_memory.rs
tbrezot: no because this will replace the trait documentation. Plus these are implementation details and have no use in the API doc. You can either document your script (which you have extracted as an independent constant) or write them as comment (for the developer's attention)

src/redis_memory.rs
tbrezot: You may be right, I didn't notice the types weren't the same. There may be a solution using dynamic dispatch but it is ugly. Keep it as is.

PR #92, opened 2024-11-05
Test/cloudproof v7

HatemMn:


src/adt.rs
```diff
@@ -140,4 +322,151 @@ pub(crate) mod tests {
...
+                .map(|vals| {
+                    let vals = vals.to_vec(); // [0->99] && [100->199]
+                    let mem = memory.clone(); // A reference to the same memory
+
+                    // Spawn a new task
+                    tokio::spawn(async move {
+                        for (adr, val) in vals {
+                            // All concurrent tasks will try to write to write to the same address !!!
+                            // As the operation is atomic, only one of them should succeed.
+                            // If two of them succeed, then the implementation is incorrect and guarded write is not atomic.
```
tbrezot, line 429: To my understanding, your 4 workers will try to write at addresses i*25 to (i+1)*25-1 where i is the worker ID, conditionally to the fact that no value is stored at the default address which should be 0 since you are using `u8` as address type. If the order in which the memory receives the worker's message by decreasing worker ID then all workers should succeed. You will only see one writing if the memory received the message by increasing worker ID... why should be what happens most of the time since you only have 4 workers, the first one to be launched is the ID 0, and you are testing an in-memory Memory.

Try reverse the iterator, see if I'm wrong.

src/adt.rs
```diff
@@ -140,4 +322,151 @@ pub(crate) mod tests {
...
+            assert_eq!(
+                res.iter().filter(|&&x| x.is_some()).count(),
+                1,
+                "Only one value should be Some(...), the others should be None. Please check the implementation of guarded_write performs the write operation atomically."
+            );
+            assert_eq!(
+                res.iter().filter(|&&x| x.is_none()).count(),
+                res.len() - 1,
+                "All values but one should be equal to None. Please check the implementation of guarded_write performs the write operation atomically."
+            );
```
tbrezot, line 457: useless check since an option only has two variants.

src/adt.rs
```diff
@@ -140,4 +322,151 @@ pub(crate) mod tests {
...
+            assert_ne!(
+                old, res,
+                "The memory state did not change, meaning no write operation was performed."
+            );
+            assert_eq!(res.len(), n, "The memory state should contain n values.");
+            assert_eq!(
+                res.iter().filter(|&&x| x.is_some()).count(),
+                1,
+                "Only one value should be Some(...), the others should be None. Please check the implementation of guarded_write performs the write operation atomically."
+            );
```
tbrezot, line 452: does this really pass?

src/adt.rs
```diff
@@ -140,4 +322,151 @@ pub(crate) mod tests {
             assert_eq!(res, values);
         }
     }
+
+    mod memory_tests {
```
tbrezot, line 326: what are you doing here? You seem to be testing an object that is not used...

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "rust-analyzer.cargo.features": ["in_memory"]
+    "rust-analyzer.cargo.features": ["test"]
```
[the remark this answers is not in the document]
tbrezot: please remove this file from the repository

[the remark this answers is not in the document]
HatemMn: Okay deleted it and added a line in .gitignore to never commit again

Cargo.toml
```diff
@@ -5,12 +5,14 @@ edition = "2021"
...
 [features]
 bench = []
+test = []
 
 [dependencies]
 aes = "0.8.4"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
 rand_core = "0.6.4"
+redis = {version = "0.27.5"}
```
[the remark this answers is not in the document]
HatemMn: yes true

src/address.rs
```diff
@@ -61,6 +63,21 @@ impl<const LENGTH: usize> Add<u64> for Address<LENGTH> {
...
+
+impl<const N: usize> ToRedisArgs for Address<N> {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + RedisWrite,
+    {
+        out.write_arg(self.as_ref());
+    }
+}
+
```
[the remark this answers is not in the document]
HatemMn: gated it behind cloudproof

src/adt.rs
```diff
@@ -67,23 +67,205 @@ pub trait MemoryADT {
...
     /// Write the given words at the given addresses if the word currently stored at the guard
     /// address is the given one, and returns this guard word.
     fn guarded_write(
         &self,
         guard: (Self::Address, Option<Self::Word>),
         tasks: Vec<(Self::Address, Self::Word)>,
-    ) -> impl Send + Sync + Future<Output = Result<Option<Self::Word>, Self::Error>>;
+    ) -> impl Send + Future<Output = Result<Option<Self::Word>, Self::Error>>;
+}
+// use futures::executor::block_on;
```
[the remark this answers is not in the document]
HatemMn: this is outdated ? I don't understand. If its outdated lets mark it as resolved as i see no ded code

src/adt.rs
```diff
@@ -67,23 +67,205 @@ pub trait MemoryADT {
...
+
+// ! This module defines tests any implementation of the MemoryADT interface must pass.
+#[cfg(feature = "test")]
+pub mod memory_tests {
+    use super::MemoryADT;
+
+    /// Runs all sequential tests
+
+    pub async fn test_batch_read_empty<T>(memory: &T)
+    where
```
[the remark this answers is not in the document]
HatemMn: done

src/adt.rs
```diff
@@ -67,23 +67,205 @@ pub trait MemoryADT {
...
+    }
+
+    pub async fn test_single_write_and_read<T>(memory: &T)
+    where
+        T: MemoryADT,
+        T::Address: std::fmt::Debug + PartialEq + Default,
+        T::Word: std::fmt::Debug + PartialEq + Default,
+        T::Error: std::fmt::Debug,
+    {
+        // Write a single value
```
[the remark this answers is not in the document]
HatemMn: I didn't really understand the comment sorry
```
        let write_result = memory
            .guarded_write(
                (T::Address::default(), None),
                vec![(T::Address::default(), T::Word::default())],
```

The guard address ( "default" ) is the same as the address where i will write, which means it is (in an implicit manner) asserted that its empty, no ?
If not empty the next assert 
```
        assert_eq!(write_result, None);

```

will fail

src/adt.rs
HatemMn: separate vector adt tests from memory adt local tests

src/in_memory_store.rs
```diff
@@ -92,46 +92,15 @@ impl<Address: Hash + Eq + Debug + Clone, Value: Clone + Eq + Debug> IntoIterator
     }
 }
 
+#[cfg(feature = "test-utils")]
 #[cfg(test)]
```
tbrezot, line 96: Something like that is better:

```suggestion
#[any(cfg(test), cfg(feature = "test-utils")]
```

src/test/memory.rs
```diff
@@ -58,81 +58,121 @@ where
...
         const N: usize = 1000; // number of threads
         let mut rng = StdRng::from_seed(seed);
-        let guard_address = rng.gen::<u8>(); // address used as guard
+        let counter_addr = rng.gen::<u128>(); // Random address for a counter
+
+        // Initialize counter to 0
+        memory
+            .guarded_write((counter_addr, None), vec![(counter_addr, 0)])
+            .await
+            .unwrap();
```
tbrezot, line 135: Do not initialize here, let concurrent worker try initializing it.

src/test/memory.rs
```diff
@@ -58,81 +58,121 @@ where
...
-                    // All concurrent tasks will try to write to the same address
-                    // As the operation is atomic, only one of them should succeed and return true.
-                    mem.guarded_write((adr, None), vec![(guard_address, 1)])
-                        .await
-                        .unwrap()
-                        .is_none()
+                    loop {
+                        // Read current value
+                        let current = mem.batch_read(vec![counter_addr]).await.unwrap()[0]
+                            .expect("Counter should exist");
```
tbrezot, line 144: No need for this: first try with `None`, then use the value returned by the previous `gwrite`.

src/test/mod.rs
```diff
@@ -1 +1,2 @@
+#[cfg(feature = "test-utils")]
```
tbrezot, line 1: No need for that, it is already gated in `lib.rs`.

src/lib.rs
```diff
@@ -9,18 +9,25 @@ mod findex;
 mod ovec;
 mod secret;
 mod symmetric_key;
+#[cfg(any(test, feature = "bench", feature = "test-utils"))]
+mod test;
 mod value;
 
 pub use address::Address;
 pub use adt::{IndexADT, MemoryADT};
+pub use encoding::{dummy_decode, dummy_encode};
```
tbrezot, line 18: I am not sure we want to publicly expose these, as they are really rudimentary...

src/test/memory.rs
HatemMn: done

src/test/memory.rs
HatemMn: done

src/test/mod.rs
HatemMn: true done

src/lib.rs
HatemMn: exposing these was the original goal of this PR as there is no clear way of instantiation in cloudproof rust without them

HatemMn:
Follow up : https://github.com/Cosmian/findex/pull/93

PR #97, opened 2024-12-03
fix: cleanup deny.toml and delete docker-compose.yml

Manuthor:


HatemMn (changes requested):
No problems on my end

I want to just point it this recurrent issue to maybe fix it for good here : 
image

As we said earlier this command runs it fine `cargo +nightly fmt --all `
image

After reconsideration I think it's better to setup either for stable OR nightly channel for both compiling and formatting, this will make workflows less ambigious, what do you think ?

I sollicite @tbrezot opinion too on the matter

Manuthor:
> No problems on my end
> 
> I want to just point it this recurrent issue to maybe fix it for good here : image
> 
> As we said earlier this command runs it fine `cargo +nightly fmt --all ` image
> 
> After reconsideration I think it's better to setup either for stable OR nightly channel for both compiling and formatting, this will make workflows less ambigious, what do you think ?
> 
> I sollicite @tbrezot opinion too on the matter

Let's remove the `.rustfmt.toml`. 
IMHO, formatting can't drive the choice of the toolchain version.

tbrezot (approved):

tbrezot:
agrees with @Manuthor here: toolchain must be stable, and formatting requirements getting one the way must be removed

HatemMn (approved):
@Manuthor I let you delete the file as you and theophile said, you can directly rebase the branch after that

PR #99, opened 2024-12-11
update key length

tbrezot:


tbrezot:
Formatting diffs are due to bump of toolchain version, sorry

PR #112, opened 2025-01-15
Feat/findex server v7

HatemMn:


src/adt.rs
```diff
@@ -23,14 +23,14 @@ pub trait IndexADT<Keyword: Send + Sync + Hash, Value: Send + Sync + Hash> {
         &self,
         kw: Keyword,
         bindings: impl Sync + Send + IntoIterator<Item = Value>,
-    ) -> impl Send + Sync + Future<Output = Result<(), Self::Error>>;
+    ) -> impl Send + Future<Output = Result<(), Self::Error>>;
```
HatemMn, line 26: In a previous time, in the previous updates, we encountered issues on findex server that were solved by taking out this Sync

Now then, after reconsideration, I do beleive it's a good idea to take it out for the following reason @tbrezot / @Manuthor  please give me your opinions on the matter


After research, I understand the following : for some type T
- implementing Sync means that &T is send. In other word, we can have some &T variables
- implementing Future<...> means the poll method is implemented, the signature of the poll is the following 

```rust
fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output>;
```
In other terms, we need a &mut T otherwise the poll is impossible

Now the problem is that rust type system othorizes us to have either mutiple &T or only one &mut T  BUT not the two at the same time, we hence have a problem here

I cite some comment I saw on the internet : 

>  https://docs.rs/tokio/latest/tokio/task/fn.spawn.html
> 
> Notice how everything is Send + 'static but Sync is nowhere to be seen.

src/memory/redis_store.rs
```diff
@@ -0,0 +1,173 @@
...
+            cmd.arg("false")
+        };
+
+        let cmd = bindings
+            .iter()
+            .fold(cmd.arg(bindings.len()), |cmd, (a, w)| cmd.arg(&**a).arg(w));
+
+        cmd.query_async(&mut self.manager.clone())
+            .await
+            .map_err(std::convert::Into::into)
```
tbrezot, line 118: Something like that is clearer:
```suggestion
            .map_err(Error::from)
```

src/test/memory.rs
```diff
@@ -0,0 +1,172 @@
+// ! This module defines tests any implementation of the MemoryADT interface
+// must pass.
```
tbrezot, line 2: then why isn't it in `adt.rs`? Also, these tests need to be exported to allow other libraries implementing a memory to use them.

src/test/memory.rs
```diff
@@ -0,0 +1,172 @@
...
+    let expected_result = vec![None, None, None];
+    assert_eq!(
+        empty_read_result, expected_result,
+        "Test batch_read of empty addresses failed.\nExpected result : {expected_result:?}. Got : \
+         {empty_read_result:?}. Seed : {seed:?}"
+    );
+
+    // Generate a random address and a random word that we save
+    let random_address = rng.gen::<u128>().to_be_bytes();
+    let random_word = rng.gen::<u128>().to_be_bytes();
```
tbrezot, line 34: Proceed with the conversion here to simplify later uses.

src/test/memory.rs
```diff
@@ -0,0 +1,172 @@
+// ! This module defines tests any implementation of the MemoryADT interface
+// must pass.
+use rand::{Rng, SeedableRng, rngs::StdRng};
+
+use crate::MemoryADT;
+
+pub async fn test_single_write_and_read<T>(memory: &T, seed: [u8; 32])
```
tbrezot, line 7: Document the testing strategy here instead of inside the body of the function.

src/test/memory.rs
```diff
@@ -0,0 +1,172 @@
...
+        for handle in handles {
+            handle.join().unwrap().await;
+        }
+
+        let final_count = memory.batch_read(vec![a.into()]).await.unwrap()[0]
+            .clone()
+            .expect("Counter should exist");
+
+        assert_eq!(
+            u128::from_be_bytes(final_count.clone().into()),
```
tbrezot, line 163: detail but unneeded clone

Cargo.toml
```diff
@@ -29,13 +30,24 @@ rand_chacha = "0.3.1"
 rand_core = "0.6.4"
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 xts-mode = "0.5.1"
-zeroize = { version = "1.8.1", features = ["derive"] }
+zeroize = { version = "1.8", features = ["derive"] }
+redis = { version = "0.28.1", features = [
+    "aio",
+    "tokio-comp",
```
tbrezot, line 36: Isn't this only required for testing only?

src/memory/redis_store.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let cmd = if let Some(byte_array) = guard_value {
+            cmd.arg(&byte_array)
+        } else {
+            cmd.arg("false")
+        };
+
+        let cmd = bindings
+            .iter()
+            .fold(cmd.arg(bindings.len()), |cmd, (a, w)| cmd.arg(&**a).arg(w));
```
tbrezot, line 114: This syntax is good: it doesn't require addresses to implement anything. Maybe reuse it in `batch_read`?

tbrezot (commented):
Looking good. Some small things to fix though.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.83.0"
+components = ["rustfmt"]
```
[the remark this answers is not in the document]
Manuthor: yes we really need to freeze the version of the toolchain since `rustfmt` will evolve in time in `stable` channel.

src/memory/redis_store.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    use super::*;
+    use crate::test::memory::{
+        test_guarded_write_concurrent, test_single_write_and_read, test_wrong_guard,
+    };
+
+    fn get_redis_url() -> String {
+        std::env::var("REDIS_HOST").map_or_else(
+            |_| "redis://localhost:9999".to_owned(),
+            |var_env| format!("redis://{var_env}:9999"),
```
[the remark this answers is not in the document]
Manuthor: by default, Redis should listen on `6379` (this is the default port)

src/lib.rs
```diff
@@ -9,18 +9,22 @@ mod memory;
 mod ovec;
 mod secret;
 mod symmetric_key;
+#[cfg(any(test, feature = "test-utils"))]
```
Manuthor, line 12: Why do we need a `test-utils` feature?

src/memory/redis_store.rs
```diff
@@ -0,0 +1,177 @@
...
+        Self { inner: e }
+    }
+}
+
+#[derive(Clone)]
+pub struct RedisMemory<Address, Word> {
+    pub manager: ConnectionManager,
+    script_hash: String,
+    a: PhantomData<Address>,
+    w: PhantomData<Word>,
```
Manuthor, line 54: Could you remind me why we need to `PhantomData` here? (and add it to the code)

Manuthor (commented):
LGTM. Just a few questions

src/adt/interfaces.rs
```diff
@@ -23,14 +23,14 @@ pub trait IndexADT<Keyword: Send + Sync + Hash, Value: Send + Sync + Hash> {
         &self,
         kw: Keyword,
         bindings: impl Sync + Send + IntoIterator<Item = Value>,
-    ) -> impl Send + Sync + Future<Output = Result<(), Self::Error>>;
+    ) -> impl Send + Future<Output = Result<(), Self::Error>>;
```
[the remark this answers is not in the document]
tbrezot: So apparently, futures *do not need* the `Sync` trait as their ownership is always passed (they have no `&self` method).

src/lib.rs
tbrezot: Generic tests are exported should  someone implement a custom memory. The in-memory back-end could be gated behind this feature too, instead of having an additional feature `bench`.

src/memory/redis_store.rs
tbrezot: To specialize a Redis memory to its intended address and word types (generics cannot be added unless they are used, thus the use for phantom data). Otherwise the same instance could be used with implementations of the `MemoryADT` for different types, which in turns might happen if the compiler infers the wrong types.

Manuthor (dismissed):

tbrezot:
Please, merge #114 before merging this MR.

src/memory/redis_store.rs
alekece: Small tip here:
Combine all your generic type in one `PhantomData` using tuple such as:
```rust
pub struct RedisMemory<Address, Word> {
    pub manager: ConnectionManager,
    script_hash: String,
    _marker: PhantomData<(Address, Word)>,
    ...
```

HatemMn: thanks @alekece for the suggestion !

src/adt/interfaces.rs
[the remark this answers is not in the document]
HatemMn: I agree with the above statement. Shall we mark this solved then ?

PR #114, opened 2025-01-21
Small improvements.

tbrezot:
Notable changes:
- remove a clone when instantiating a Redis memory;
- strengthen the concurrency test by looping `N` times in each concurrent worker
- remove a redundant condition in the g-write script
- remove some unnecessary bounds in the generic tests
- remove redundant documentation (please, refrain your LLM)
- remove the `bench` feature

### DIFF
--- a/src/memory/redis_store.rs
+++ b/src/memory/redis_store.rs
@@ -17,7 +17,7 @@ local length = ARGV[3]
 local value = redis.call('GET',ARGV[1])
 
 -- compare the value of the guard to the currently stored value
-if((value==false) or (not(value == false) and (guard_value == value))) then
+if ((value == false) or (guard_value == value)) then
     -- guard passed, loop over bindings and insert them
     for i = 4,(length*2)+3,2
     do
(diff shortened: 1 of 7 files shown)


src/memory/redis_store.rs
```diff
@@ -48,7 +48,7 @@ impl From<RedisError> for MemoryError {
 
 #[derive(Clone)]
 pub struct RedisMemory<Address, Word> {
-    pub manager: ConnectionManager,
+    manager: ConnectionManager,
```
HatemMn, line 51: Why was this put back private ?  This modification is blocking in findex server and will stop it from working

I will try to summarize my motivation for making manager public (even tho it was not) : 
- Legacy code (findex v6) had access to the manager
- Which means legacy v6 implementations of the following two files call the manager : 

This one, permission : https://github.com/Cosmian/findex-server/blob/342ea98bca3692384df05787bd7b2507bf38c68b/crate/server/src/database/redis/permissions.rs
And this one, datasets :  https://github.com/Cosmian/findex-server/blob/b3a0026e442aeebaba07a789b8c6d220b999b9e7/crate/server/src/database/redis/datasets.rs


Both need a clone of the manager, @Manuthor as you originally authored the code, what do you think ?

My opinion is that, conceptually talking it's logic that the manager is a private field but due to the server needs we shall - at least for now - make it public or add some Getter for him to not block the code of the permissions and datasets, which itself already has other problems with the transaction implementation being disussed here : https://github.com/redis-rs/redis-rs/pull/1485

HatemMn (commented):
Looks good, the PR summary basically says everything

We just need to find a solution for the conn manager before merging to avoid killing the server

PR #118, opened 2025-01-28
Add generic encodings, and a generic test all encodings must pass.

tbrezot:


src/encoding.rs
```diff
@@ -1,78 +1,384 @@
...
+                        if n <= available::<WORD_LENGTH>(*pos) {
+                            cur_w[*pos..*pos + n].iter().for_each(|b| bytes.push(*b));
+                            *pos += n;
+                            return Some(bytes);
+                        } else {
+                            cur_w[*pos..].iter().for_each(|b| bytes.push(*b));
+                            n -= available::<WORD_LENGTH>(*pos);
+                            w = ws.next();
+                            *pos = 0;
+                        }
```
HatemMn, line 188: why not simply ?

```suggestion
                        bytes.extend_from_slice(&cur_w[pos..pos + n]);
```

copying chunk of memory feels 'lighter' than iterating even the output is exactly same

src/encoding.rs
```diff
@@ -1,78 +1,384 @@
...
+                    } else {
+                        // If there is no more words and not enough bytes could be read,
+                        // let the caller manage.
+                        return None;
```
HatemMn, line 192: same question

HatemMn, line 192: would it break the logic if we keep n (an outsider value) immutable to not mutate the caller's env, and use a local value instead ?

src/encoding.rs
```diff
@@ -1,78 +1,384 @@
...
+        };
+
+        // The position must be exposed since padding detection cannot be
+        // performed in `read_bytes`.
+        let mut pos = 0;
+        let mut vs = HashSet::<Value>::new();
+        while let Some(m0) = read_bytes(1, &mut pos) {
+            if (m0[0] >> 7) == 1 {
+                if let Some(m1) = read_bytes(1, &mut pos) {
+                    let m = <u16>::from_be_bytes([m0[0], m1[0]]);
```
HatemMn, line 205: I understand what this is for does but shouldn't it be commented it for other readers ?

IMO we should have XOR(verbose variable names like ~~b1~~ metadata_byte1 , comments )

HatemMn (commented):
LGMT :D

src/encoding.rs
```diff
@@ -1,78 +1,373 @@
...
+                    ws.push(w);
+                    bytes = &bytes[available(pos)..];
+                    w = [0; WORD_LENGTH];
+                    pos = 0;
+                }
+            }
+        };
+
+        for v in vs {
+            let metadata = encode_metadata(op, v.as_ref().len())?.to_be_bytes();
```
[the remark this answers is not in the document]
tbrezot: `<u16>::as_ref()` does not to what we want :  we want to get the bytes here.

src/encoding.rs
tbrezot: applied

src/encoding.rs
tbrezot: variable is not a mutable reference so everything is safe

src/encoding.rs
tbrezot: ok

src/encoding.rs
```diff
@@ -1,78 +1,384 @@
...
+                    }
+                } else {
+                    return Err("cannot read second metadata byte".to_string());
+                }
             } else {
-                res[0] = 0;
+                // Reading an empty byte when the first byte of some metadata
+                // was expected means the rest of the word is padding: advance
+                // the position to the end of the word.
+                pos += available::<WORD_LENGTH>(pos);
```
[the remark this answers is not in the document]
tbrezot: this line skips any internal padding

HatemMn (approved):

PR #121, opened 2025-02-12
fix: convert indexing slicing to error VectorADT::read()

Manuthor:
- fix: convert indexing slicing to error VectorADT::read()
- rename traits parameters (`kw`-> `keyword`, `a`-> `addresses`) and make them consistent with implementations
- add generic `WORD_LENGTH` on `test_utils` functions (functions that are actually used on HTTP client & KMS encryption layer implementations 
- upgrades rand crates (`rand`, `rand_chacha` and `rand_core`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@
 
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-exclude: datasets|tests_data|documentation/Findex.pdf|tests/first_names.txt
+exclude: benches/data|benches/make_figures.tex
 repos:
   - repo: https://github.com/compilerla/conventional-pre-commit
     rev: v2.1.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -118,20 +118,21 @@ repos:
         args: [--skip-string-normalization]
 
   - repo: https://github.com/Cosmian/git-hooks.git
-    rev: v1.0.16
+    rev: v1.0.36
     hooks:
       - id: stable-cargo-format
-      - id: cargo-upgrade
-      - id: cargo-update
+      # - id: dprint-toml-fix
+      # - id: cargo-upgrade
+      # - id: cargo-update
       - id: cargo-machete
-      - id: cargo-outdated
-      - id: cargo-udeps
+      - id: redis-container-up
       - id: cargo-tests-all
       - id: cargo-test-doc
-      - id: clippy-autofix-all
-      - id: clippy-autofix-pedantic
-      - id: clippy-autofix-others
+      - id: clippy-autofix-all-targets-all-features
+      - id: clippy-autofix-all-targets
       - id: clippy-all-targets-all-features
+      - id: clippy-all-targets
       - id: stable-cargo-format
       - id: cargo-dry-publish
         args: [--allow-dirty]
+      - id: redis-container-down
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -8,7 +8,7 @@ use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use futures::{executor::block_on, future::join_all};
 use lazy_static::lazy_static;
 use rand_chacha::ChaChaRng;
-use rand_core::{CryptoRngCore, RngCore, SeedableRng};
+use rand_core::{CryptoRng, RngCore, SeedableRng};
 
 lazy_static! {
     static ref scale: Vec<f32> = make_scale(0, 4, 20);
@@ -23,10 +23,9 @@ fn make_scale(start: usize, stop: usize, n: usize) -> Vec<f32> {
     points
 }
 
-/// Builds an index that associates each `kw_i` to x values, both random 64-bit values.
-fn build_benchmarking_bindings_index(
-    rng: &mut impl CryptoRngCore,
-) -> Vec<([u8; 8], HashSet<[u8; 8]>)> {
+/// Builds an index that associates each `kw_i` to x values, both random 64-bit
+/// values.
+fn build_benchmarking_bindings_index(rng: &mut impl CryptoRng) -> Vec<([u8; 8], HashSet<[u8; 8]>)> {
     scale
         .iter()
         .map(|i| {
@@ -39,10 +38,9 @@ fn build_benchmarking_bindings_index(
         .collect()
 }
 
-/// Builds an index that associates 10^3 `kw_i` to a single value, both random 64-bit values.
-fn build_benchmarking_keywords_index(
-    rng: &mut impl CryptoRngCore,
-) -> Vec<([u8; 8], HashSet<[u8; 8]>)> {
+/// Builds an index that associates 10^3 `kw_i` to a single value, both random
+/// 64-bit values.
+fn build_benchmarking_keywords_index(rng: &mut impl CryptoRng) -> Vec<([u8; 8], HashSet<[u8; 8]>)> {
     (0..10usize.pow(3))
         .map(|_| {
             let kw = rng.next_u64().to_be_bytes();
@@ -53,7 +51,7 @@ fn build_benchmarking_keywords_index(
 }
 
 fn bench_search_multiple_bindings(c: &mut Criterion) {
-    let mut rng = ChaChaRng::from_entropy();
+    let mut rng = ChaChaRng::from_os_rng();
     let seed = Secret::random(&mut rng);
     let ctx_memory = MemoryEncryptionLayer::new(&seed, InMemory::default());
     let index = build_benchmarking_bindings_index(&mut rng);
@@ -80,7 +78,7 @@ fn bench_search_multiple_bindings(c: &mut Criterion) {
 }
 
 fn bench_search_multiple_keywords(c: &mut Criterion) {
-    let mut rng = ChaChaRng::from_entropy();
+    let mut rng = ChaChaRng::from_os_rng();
 
     let index = build_benchmarking_keywords_index(&mut rng);
 
@@ -104,8 +102,8 @@ fn bench_search_multiple_keywords(c: &mut Criterion) {
         for i in scale.iter() {
             let n = 10f32.powf(*i).ceil() as usize;
             group.bench_function(BenchmarkId::from_parameter(n), |b| {
-                // Attempts to bench all external costs (somehow, cloning the keywords impacts the
-                // benches).
+                // Attempts to bench all external costs (somehow, cloning the keywords impacts
+                // the benches).
                 b.iter_batched(
                     || {
                         ptx_memory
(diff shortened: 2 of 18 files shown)


benches/benches.rs
```diff
@@ -53,7 +51,7 @@ fn build_benchmarking_keywords_index(
 }
 
 fn bench_search_multiple_bindings(c: &mut Criterion) {
-    let mut rng = ChaChaRng::from_entropy();
+    let mut rng = ChaChaRng::from_os_rng();
```
tbrezot, line 54: For my information, what is the rational behind this change ?

src/adt/test_utils.rs
```diff
@@ -113,43 +117,59 @@ where
...
 
+fn u128_to_array<const WORD_LENGTH: usize>(u: u128) -> [u8; WORD_LENGTH] {
+    let mut bytes = [0u8; WORD_LENGTH];
+    bytes[..16].copy_from_slice(&u.to_be_bytes());
+    bytes
+}
+
+fn word_to_array<const WORD_LENGTH: usize>(word: [u8; WORD_LENGTH]) -> u128 {
+    let mut bytes = [0; 16];
+    bytes.copy_from_slice(&word[..16]);
```
tbrezot, line 128: This panics if `WORD_LENGTH < 16`

src/adt/test_utils.rs
```diff
@@ -113,43 +117,59 @@ where
...
+    seed: [u8; KEY_LENGTH],
+) where
     Memory: 'static + Send + Sync + MemoryADT + Clone,
-    Memory::Address: Send + From<[u8; 16]>,
-    Memory::Word: Send + Debug + PartialEq + From<[u8; 16]> + Into<[u8; 16]> + Clone + Default,
+    Memory::Address: Send + From<[u8; ADDRESS_LENGTH]>,
+    Memory::Word:
+        Send + Debug + PartialEq + From<[u8; WORD_LENGTH]> + Into<[u8; WORD_LENGTH]> + Clone,
     Memory::Error: Send + std::error::Error,
+    [(); WORD_LENGTH]: Sized,
```
tbrezot, line 146: Why is this bound necessary ? Isn't it always the case ?

src/findex.rs
```diff
@@ -93,26 +93,28 @@ impl<
...
         let words = IVec::new(l, self.el.clone()).read().await?;
         (self.decode)(words).map_err(|e| Error::Conversion(format!("{e:?}")))
     }
 
     async fn insert(
         &self,
-        kw: Keyword,
-        vs: impl Sync + Send + IntoIterator<Item = Value>,
+        keyword: Keyword,
+        bindings: impl Sync + Send + IntoIterator<Item = Value>,
```
tbrezot, line 105: No, `vs` = `values`

src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![warn(clippy::all, clippy::nursery, clippy::cargo)]
+#![allow(clippy::multiple_crate_versions)]
```
tbrezot, line 2: Why is this necessary ?

src/memory/encryption_layer.rs
```diff
@@ -108,12 +108,12 @@ impl<
     async fn guarded_write(
         &self,
         guard: (Self::Address, Option<Self::Word>),
-        bindings: Vec<(Self::Address, Self::Word)>,
+        tasks: Vec<(Self::Address, Self::Word)>,
```
tbrezot, line 111: I prefer `binding` since `task` does not mean anything while `binding` refers to a link between a name and value. Maybe `binding` can be kept for keyword/value pairs only. But then I would still prefer `words` to `tasks`.

src/ovec.rs
```diff
@@ -177,7 +177,9 @@ where
             .map_err(|e| Error::Memory(e.to_string()))?;
 
         let second_batch = {
-            let cur_header = first_batch[0]
+            let cur_header = first_batch
+                .first()
+                .ok_or_else(|| Error::MissingValue(self.a.clone(), 0))?
```
tbrezot, line 182: This is a hard error. In theory it should never happen when a memory that correctly implements the `MemoryADT` trait is used. I would therefore use a more explicit error message.

src/adt/test_utils.rs
Manuthor: indeed, removed.

src/findex.rs
Manuthor: right. Do you agree to also change it in `IndexADT` trait?

src/lib.rs
Manuthor: I didn't want to write the reason in the lib.rs (to not add unnecessary noise).
The error:
```
error: multiple versions for dependency `zerocopy`: 0.7.35, 0.8.18
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions
  = note: `-D clippy::multiple-crate-versions` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(clippy::multiple_crate_versions)]`

error: multiple versions for dependency `zerocopy-derive`: 0.7.35, 0.8.18
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions
```

The crate `ppv-lite86` is still using an old version of `zerocopy`... (not upgradable)

src/memory/encryption_layer.rs
Manuthor: right. Changing it in `MemoryADT` is OK for you?

src/findex.rs
tbrezot: okay for me

src/ovec.rs
tbrezot: No, just raise a `Error::Memory` with a clear message, mayb something like "no result found after calling `bread`, this should not happen and is likely due to a bug in the underlying memory implementation"

src/adt/interfaces.rs
```diff
@@ -7,7 +7,7 @@
 
 use std::{collections::HashSet, future::Future, hash::Hash};
 
-/// An index stores *bindings*, that associate a keyword with a value. All values bound to the same
+/// An index stores *values*, that associate a keyword with a value. All values bound to the same
```
tbrezot, line 10: No, an index does store bindings.

src/adt/interfaces.rs
```diff
@@ -18,18 +18,18 @@ pub trait IndexADT<Keyword: Send + Sync + Hash, Value: Send + Sync + Hash> {
         keyword: &Keyword,
     ) -> impl Future<Output = Result<HashSet<Value>, Self::Error>>;
 
-    /// Adds the given bindings to the index.
+    /// Adds the given values to the index.
```
tbrezot, line 21:
```suggestion
    /// Binds the given values to this keyword in the index.
```

src/adt/interfaces.rs
```diff
@@ -18,18 +18,18 @@ pub trait IndexADT<Keyword: Send + Sync + Hash, Value: Send + Sync + Hash> {
...
     fn insert(
         &self,
         keyword: Keyword,
-        bindings: impl Sync + Send + IntoIterator<Item = Value>,
+        values: impl Sync + Send + IntoIterator<Item = Value>,
     ) -> impl Send + Future<Output = Result<(), Self::Error>>;
 
-    /// Removes the given bindings from the index.
+    /// Removes the given values from the index.
```
tbrezot, line 28:
```suggestion
    /// Unbinds the given values from this keyword in the index.
```

src/adt/test_utils.rs
```diff
@@ -18,6 +18,21 @@ fn gen_bytes<const BYTES_LENGTH: usize>(rng: &mut impl RngCore) -> [u8; BYTES_LE
     bytes
 }
 
+fn u128_to_array<const WORD_LENGTH: usize>(u: u128) -> [u8; WORD_LENGTH] {
+    let mut bytes = [0u8; WORD_LENGTH];
+    bytes[..16].copy_from_slice(&u.to_be_bytes());
+    bytes
```
tbrezot, line 24: panics if `WORD_LENGTH < 16`.

src/findex.rs
```diff
@@ -93,26 +93,28 @@ impl<
...
+        self.push(Op::Insert, keyword, bindings.into_iter().collect())
+            .await
     }
 
     async fn delete(
         &self,
-        kw: Keyword,
-        vs: impl Sync + Send + IntoIterator<Item = Value>,
+        keyword: Keyword,
+        bindings: impl Sync + Send + IntoIterator<Item = Value>,
```
HatemMn: maybe `values` ?

HatemMn (approved):
Looks good, most important details are in TBZ comments

src/memory/encryption_layer.rs
tbrezot: > Tasks is an idiomatic name for this kind of server operation

"server" has no meaning here, this is not the correct abstraction.

> I still have trouble when I read "bindings", often not knowing it binds what to what

Well to me `binding: (X, Y)` binds a value of type `X` to a value of type `Y`. That was the rational behind this name.

HatemMn: I think I have trouble with the name "bindings" here because I have a more "macro" look, reading the function signature, I expect a Vector of "stuff" to be worked on

If for example the function has as signature only an adr/word pair, I'd say binding looks fit

In all, this is a highly subjective matter, I think that in the end any of the 3 proposed names will be a good fit

src/memory/encryption_layer.rs
```diff
@@ -108,12 +108,12 @@ impl<
...
     ) -> Result<Option<Self::Word>, Self::Error> {
-        let (a, v) = guard;
-        let tok = self.permute(a);
+        let (address, v) = guard;
```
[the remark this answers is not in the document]
Manuthor: done in 42a347ea9cebd6392499c8933cc7de869946bcea

src/adt/test_utils.rs
Manuthor: done in 42a347ea9cebd6392499c8933cc7de869946bcea

src/ovec.rs
Manuthor: done in 42a347ea9cebd6392499c8933cc7de869946bcea

.pre-commit-config.yaml
```diff
@@ -118,20 +118,21 @@ repos:
...
   - repo: https://github.com/Cosmian/git-hooks.git
-    rev: v1.0.16
+    rev: v1.0.36
     hooks:
       - id: stable-cargo-format
-      - id: cargo-upgrade
-      - id: cargo-update
+      # - id: dprint-toml-fix
+      # - id: cargo-upgrade
+      # - id: cargo-update
```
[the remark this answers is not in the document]
Manuthor: made them `manual` (to keep and run them from time to time)

src/lib.rs
Manuthor: ## Who Depends on `ppv-lite86`?

- `rand_chacha` v0.9.0 depends on `ppv-lite86` v0.2.20, which depends on `zerocopy` v0.7.35.

## And Who Depends on a More Recent Version of `zerocopy`?

- `rand` v0.9.0 and `rand_core` v0.9.1.

I believe the maintainers of `ppv-lite86` did not correctly publish crate v0.2.20, as the source code actually depends on `zerocopy` v0.8, contrary to the published crate:  
[https://crates.io/crates/ppv-lite86/0.2.20/dependencies](https://crates.io/crates/ppv-lite86/0.2.20/dependencies), which depends on v0.7.35.

tbrezot (approved):

PR #122, opened 2025-02-19
SQL

HatemMn:
## Overview
Adds an sqlite memory to be used with Findex SSE like the already existing `in_memory` and `redis` interfaces

Uses the [rusqlite library](https://github.com/rusqlite/rusqlite) , which is essentially some Rust wrappers around the native SQLite C library. In other words **we have direct, low-level access to the SQLite database file on disk** , no deamons nor ORMs.

## Technical considerations : 
- You might notice the usage of a pool. To explain it briefly, using a classic sqlite `Connection` as-is is not possible because it contains a RefCell [and RefCell is never sync](https://doc.rust-lang.org/std/cell/struct.RefCell.html#impl-Sync) while the signature of memory ADT requires a reference ( &self ).
A solution to this would be to use Mutex ... But actually as we do know pools are almost always more efficcient, so by using pool, we solve two issues with 1 stone
- Retries on transaction failure : is it highly unlikely for a transaction to even fail, but in case of this happenning, an exponential backoff retry loop was added and is optionally activated. If not explicitly setup, it won't be used as it can slow programs that have low contention

## Refs :
-   https://github.com/ivanceras/r2d2-sqlite/blob/3e2db834b87e6d5fcc5316e67742b34ce6fcff0d/tests/test.rs
- https://en.wikipedia.org/wiki/Exponential_backoff


HatemMn:
moved https://github.com/Cosmian/findex/pull/125

PR #125, opened 2025-03-03
Feat/sqlite3

HatemMn:
## Overview
Adds an sqlite memory to be used with Findex SSE like the already existing `in_memory` and `redis` interfaces

Uses the [rusqlite library](https://github.com/rusqlite/rusqlite) , which is essentially some Rust wrappers around the native SQLite C library. In other words **we have direct, low-level access to the SQLite database file on disk** , no deamons nor ORMs.

## Technical considerations : 
- You might notice the usage of a pool. To explain it briefly, using a classic sqlite `Connection` as-is is not possible because it contains a RefCell [and RefCell is never sync](https://doc.rust-lang.org/std/cell/struct.RefCell.html#impl-Sync) while the signature of memory ADT requires a reference ( &self ).
A solution to this would be to use Mutex ... But actually as we do know pools are almost always more efficcient, so by using pool, we solve two issues with 1 stone
- Retries on transaction failure : is it highly unlikely for a transaction to even fail, but in case of this happenning, an exponential backoff retry loop was added and is optionally activated. If not explicitly setup, it won't be used as it can slow programs that have low contention

## Refs :
-   https://github.com/ivanceras/r2d2-sqlite/blob/3e2db834b87e6d5fcc5316e67742b34ce6fcff0d/tests/test.rs
- https://en.wikipedia.org/wiki/Exponential_backoff


src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        pool.get().unwrap().execute(CREATE_TABLE_SCRIPT, [])?;
+        Ok(Self {
+            pool,
+            exponential_backoff: backoff.unwrap_or(false),
+            _marker: PhantomData,
+        })
+    }
+
+    /// Connects to a known DB url using the given URL.
+    pub fn connect(url: &str, backoff: Option<bool>) -> Result<Self, SqlMemoryError> {
```
tbrezot, line 47: URL is misleading:
```suggestion
    /// Connects to a known DB url using the given path.
    pub fn connect(path: &impl AsRef<Path>, backoff: Option<bool>) -> Result<Self, SqlMemoryError> {
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        // Using a transaction to ensure the read is atomic
+        let tx = conn.transaction()?;
+
+        let query = format!(
+            "SELECT adr, word FROM findex_aw_store WHERE adr IN ({})",
+            (0..addresses.len())
+                .map(|_| "?")
+                .collect::<Vec<&str>>()
+                .join(",")
+        );
```
tbrezot, line 89: No need for a transaction here since the query is a single select.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+    // only enable this in case of a high contention scenario
+    exponential_backoff: bool,
+    _marker: PhantomData<(Address, Word)>,
+}
+
+const MAX_POOL_SIZE: u32 = 10; // placeholder
+const CREATE_TABLE_SCRIPT: &str = "
+CREATE TABLE IF NOT EXISTS findex_aw_store
+(
+    adr BLOB PRIMARY KEY,
```
tbrezot, line 28: please no spelling upon abbreviating: either use `(a, w)` as usual or `(addr, word)` if you really want to innovate.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        let query = format!(
+            "SELECT adr, word FROM findex_aw_store WHERE adr IN ({})",
+            (0..addresses.len())
+                .map(|_| "?")
+                .collect::<Vec<&str>>()
+                .join(",")
+        );
+
+        // Convert addresses to SQL parameters
+        let params: Vec<ToSqlOutput> = addresses
```
tbrezot, line 92: Why is the type `ToSqlOutput` when your addresses are actually *inputs*?

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+                rusqlite::types::ToSqlOutput::Borrowed(rusqlite::types::ValueRef::Blob(&**addr))
+            })
+            .collect();
+
+        // Execute the query
+        let mut stmt = tx.prepare(&query)?;
+        let mut rows = stmt.query(rusqlite::params_from_iter(params))?;
+
+        // Process all rows
+        let mut result_map = std::collections::HashMap::new();
```
tbrezot, line 104: Using a map here is just wasting CPU cycles: use a vector instead (SELECT should return values in the correct order).

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        while let Some(row) = rows.next()? {
```
tbrezot, line 105: Using a `for` here would prove you can read back the correct number of values.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+            let adr: Self::Address = row.get::<_, [u8; ADDRESS_LENGTH]>(0)?.into();
+            let word: Self::Word = row.get(1)?; // tb verified
+            // Convert to fixed-size array
```
tbrezot, line 108: erroneous comment
```suggestion
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        // Execute the query
+        let mut stmt = tx.prepare(&query)?;
+        let mut rows = stmt.query(rusqlite::params_from_iter(params))?;
+
+        // Process all rows
+        let mut result_map = std::collections::HashMap::new();
+        while let Some(row) = rows.next()? {
+            let adr: Self::Address = row.get::<_, [u8; ADDRESS_LENGTH]>(0)?.into();
+            let word: Self::Word = row.get(1)?; // tb verified
```
tbrezot, line 107: no TODO like that in PR please: either leave an understandable message or fix it.
```suggestion
            let word: Self::Word = row.get(1)?;
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+            // Convert to fixed-size array
+            result_map.insert(adr, word);
+        }
+
+        // Create results in the same order as input addresses
+        // TODO: Check if this is necessary of if the order is always respected anyway
+        let results = addresses
+            .iter()
+            .map(|addr| result_map.get(addr).copied())
+            .collect();
```
tbrezot, line 117: Fix this TODO. Moreover you are wasting time here with unneeded copies.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    async fn guarded_write(
+        &self,
+        guard: (Self::Address, Option<Self::Word>),
+        bindings: Vec<(Self::Address, Self::Word)>,
+    ) -> Result<Option<Self::Word>, Self::Error> {
+        let mut conn = self.pool.get()?;
+
+        let (guard_address, guard_value) = guard;
```
tbrezot, line 129: This is not a value but a word.
```suggestion
        let (ag, wg) = guard;
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        let max_retries = 5;
```
tbrezot, line 131: magic number

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        // start with a very small delay since SQLite is used locally
+        // only used if exponential_backoff is enabled
+        let base_delay_ms = 2;
```
tbrezot, line 134: Magic number once again.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        let mut retries = 0;
+        let mut last_err = rusqlite::Error::ExecuteReturnedResults;
+
+        while retries < max_retries {
```
tbrezot, line 138: Use a `for` instead! Only use `while` in case it is highly more idiomatic of you cannot statically determine the bound.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+            if self.exponential_backoff && retries > 0 {
+                sleep(Duration::from_millis(base_delay_ms * 2_u64.pow(retries)));
+            }
```
tbrezot, line 141: no need for the 0 special case

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+            retries += 1;
+            let tx = conn.transaction()?;
+            let first_query = "SELECT word AS previous_value FROM findex_aw_store WHERE adr = ?";
+            // Convert addresses to SQL parameters
```
tbrezot, line 145: wrong comment !
```suggestion
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+                // No rows returned, the address is not in the database
+                Err(rusqlite::Error::QueryReturnedNoRows) => None,
+                Err(e) => {
+                    tx.rollback()?;
+                    last_err = e;
+                    continue; // Continue retry loop
+                }
+            };
+
+            if previous_word == guard_value {
```
tbrezot, line 161: `previous_word` is ill-named: it may be the `current_word` and is named `w_g^*` in the paper.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+                [rusqlite::types::Value::Blob(guard_address.to_vec())],
+                |row| row.get::<_, Self::Word>(0),
+            ) {
+                Ok(word) => Some(word),
+                // No rows returned, the address is not in the database
+                Err(rusqlite::Error::QueryReturnedNoRows) => None,
+                Err(e) => {
+                    tx.rollback()?;
+                    last_err = e;
+                    continue; // Continue retry loop
```
tbrezot, line 157: If you need comments like that, it means the code is too complex and that is may be time to break it into smaller units.
```suggestion
                    continue;
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+            };
+
+            if previous_word == guard_value {
+                // guard passed, flatten bindings
+                let params2 = params_from_iter(bindings.iter().flat_map(|(a, w)| {
+                    vec![
+                        rusqlite::types::Value::Blob(a.to_vec()),
+                        rusqlite::types::Value::Blob(w.to_vec()),
+                    ]
+                }));
```
tbrezot, line 168: Are you sure you need to copy everything?

tbrezot: Also `params2` is poorly named.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+                let pts_dinterrogation = (0..bindings.len())
```
tbrezot, line 169: seriously?

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,225 @@
...
+        let mut retries = 0;
+        let mut last_err = rusqlite::Error::ExecuteReturnedResults;
+
+        while retries < max_retries {
+            if self.exponential_backoff && retries > 0 {
+                sleep(Duration::from_millis(base_delay_ms * 2_u64.pow(retries)));
+            }
+            retries += 1;
+            let tx = conn.transaction()?;
+            let first_query = "SELECT word AS previous_value FROM findex_aw_store WHERE adr = ?";
```
tbrezot, line 144: no need for aliasing:
```suggestion
            let first_query = "SELECT word FROM findex_aw_store WHERE adr = ?";
```

HatemMn (commented):
INFO : explicit busy handler was delete

HatemMn:
Next step : trying async ?

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,215 @@
...
+
+const CREATE_TABLE_SCRIPT: &str = "PRAGMA synchronous = NORMAL;
+CREATE TABLE IF NOT EXISTS memory (
+    a BLOB PRIMARY KEY,
+    w BLOB NOT NULL
+);";
+
+impl<Address, Word> SqliteMemory<Address, Word> {
+    /// Connects to a known DB using the given path.
+    pub async fn connect(path: &str) -> Result<Self, SqliteMemoryError> {
```
HatemMn, line 74: (this is a standalone comment for reviewers)

I restored the `path` to an str because : 
image
****

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - The path to the sqlite3 database file.
+    /// * `timeout` - Optional : the timeout for the busy handler in milliseconds, to be only changed if necessary.
+    pub async fn connect(path: &str, timeout: Option<Duration>) -> Result<Self, SqliteMemoryError> {
+        // This pool connections number defaults to the number of logical CPUs of the current system.
+        // The following settings are used to improve performance:
+        // - journal_mode = WAL : WAL journaling is faster than the default DELETE mode without compromising reliability
+        // - synchronous = NORMAL : makes calling the (expensive) `fsync` only done in critical moments
```
tbrezot, line 73: I see no mention of this in the code.
```suggestion
```

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+{
+    type Address = Address<ADDRESS_LENGTH>;
+    type Error = SqliteMemoryError;
+    type Word = [u8; WORD_LENGTH];
+
+    async fn batch_read(
+        &self,
+        addresses: Vec<Self::Address>,
+    ) -> Result<Vec<Option<Self::Word>>, Self::Error> {
+        // Read locks are no bottleneck, so no need to edit the default busy handler.
```
tbrezot, line 107: to be removed when removing the timeout.

tbrezot (commented):
Good job with finding the `WAL` option. If there is not concurrency issue, I would suggest removing the timeout to simplify code.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub fn in_memory() -> Result<Self, SqliteMemoryError> {
+        let pool = r2d2::Pool::builder()
+            .max_size(Self::get_cpu_count())
+            .build(SqliteConnectionManager::memory())?;
+        pool.get().unwrap().execute(CREATE_TABLE_SCRIPT, [])?;
+        Ok(Self {
+            pool,
+            _marker: PhantomData,
+        })
+    }
```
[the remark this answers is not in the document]
HatemMn: this is old code

i got rid of it because the async library is not good with in_memory DB and causes a lot of errors anyway

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+impl From<async_sqlite::Error> for SqliteMemoryError {
+    fn from(err: async_sqlite::Error) -> Self {
+        Self::AsyncSqliteError(err)
+    }
+}
+
+#[derive(Clone)]
+pub struct SqliteMemory<Address, Word> {
+    pool: Pool,
+    timeout: Duration,
```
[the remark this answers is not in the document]
HatemMn: in our last discussion we rather agreed to add it no ?

Cargo.toml
```diff
@@ -22,10 +22,12 @@ path = "src/lib.rs"
```
HatemMn, line 22: begin review

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+    type Address = Address<ADDRESS_LENGTH>;
+    type Error = SqliteMemoryError;
+    type Word = [u8; WORD_LENGTH];
+
+    async fn batch_read(
+        &self,
+        addresses: Vec<Self::Address>,
+    ) -> Result<Vec<Option<Self::Word>>, Self::Error> {
+        // Read locks are no bottleneck, so no need to edit the default busy handler.
+        let res = self
```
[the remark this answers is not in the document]
HatemMn: It's possible to not assign it by adding some strong typing tweaks

I deleted it, mark solved if you are fine with the new approach

src/memory/sqlite_store.rs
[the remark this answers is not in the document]
HatemMn: in our last discussion we rather agreed to add it no ?

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+                    tx.execute(
+                        &format!(
+                            "INSERT OR REPLACE INTO memory (a, w) VALUES {}",
+                            vec!["(?,?)"; bindings.len()].join(",")
+                        ),
+                        params_from_iter(
+                            bindings
+                                .iter()
+                                // There seems to be no way to avoid cloning here.
+                                // Wrapping `x.to_vec()` and w in `Blob` can be avoided as the underlying structure is the same.
```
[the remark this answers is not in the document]
HatemMn: deleted

src/memory/sqlite_store.rs
HatemMn: This pragma is passed on the table creation - I prefer to keep this comment if you don't mind (maybe move it up if necessary) because it is not really a trivial parameter and I prefer to document why I changed it

HatemMn (commented):
I am running some "final" test to make sure we can safely delete the timouts. It takes a lot of time, I will put the result as separate comment

src/memory/sqlite_store.rs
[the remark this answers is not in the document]
tbrezot: Yes but I thought I understood that perfs were good enough now that this is not needed anymore. If it isn't, let's remove it.

src/memory/sqlite_store.rs
tbrezot: Document things were they are or in the module doc.

src/memory/sqlite_store.rs
```diff
@@ -0,0 +1,216 @@
...
+            Self::AsyncSqliteError(err) => write!(f, "async-sqlite error: {}", err),
+        }
+    }
+}
+
+impl From<async_sqlite::Error> for SqliteMemoryError {
+    fn from(err: async_sqlite::Error) -> Self {
+        Self::AsyncSqliteError(err)
+    }
+}
```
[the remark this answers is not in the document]
HatemMn: keeping it to avoid re-exporting

src/memory/sqlite_store.rs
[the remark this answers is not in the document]
HatemMn: remevod

src/memory/sqlite_store.rs
HatemMn: done

HatemMn (commented):
image

Last review

src/memory/sqlite_store.rs
```diff
@@ -126,10 +109,13 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize> MemoryADT
                 // Return order of an SQL select statement is undefined, and
                 // mismatches are ignored. A post-processing is thus needed to
                 // generate a returned value complying to the batch-read spec.
-                Ok(addresses.iter().map(|addr| bindings.remove(addr)).collect())
+                Ok(addresses
+                    .iter()
+                    .map(|addr| bindings.remove(addr))
+                    .collect::<Vec<Option<[u8; WORD_LENGTH]>>>())
```
tbrezot, line 115:
```suggestion
                Ok(addresses.iter().map(|addr| bindings.remove(addr)).collect())
```

tbrezot (approved):
one change to revert, otherwise i'm good with it

src/adt/test_utils.rs
```diff
@@ -137,77 +137,92 @@ pub async fn test_wrong_guard<const WORD_LENGTH: usize, Memory>(
...
-            "test_guarded_write_concurrent failed. Expected the counter to be at {:?}, found \
-             {:?}.\nDebug seed : {:?}.",
-            N as u128,
-            word_to_array(final_count.into()).unwrap(),
-            seed
-        );
+    let handles = (0..n)
+        .map(|_| {
+            let m = memory.clone();
+            tokio::spawn(worker(m, a))
```
tbrezot, line 207:
```suggestion
            tokio::spawn(worker(memory.clone(), a))
```

src/adt/test_utils.rs
```diff
@@ -137,77 +137,92 @@ pub async fn test_wrong_guard<const WORD_LENGTH: usize, Memory>(
...
+        })
+        .collect::<Vec<_>>();
```
tbrezot, line 209:
```suggestion
    let handles = (0..n)
        .map(|_| tokio::spawn(worker(memory.clone(), a)))
        .collect::<Vec<_>>();
```

src/adt/test_utils.rs
```diff
@@ -137,77 +137,92 @@ pub async fn test_wrong_guard<const WORD_LENGTH: usize, Memory>(
...
+///
+/// # Arguments
+///
+/// * `memory` - The Memory ADT implementation to test.
+/// * `seed` - The seed used to initialize the random number generator.
+/// * `n_threads` - The number of threads to spawn. If None, defaults to 100.
 pub async fn test_guarded_write_concurrent<const WORD_LENGTH: usize, Memory>(
     memory: &Memory,
     seed: [u8; KEY_LENGTH],
+    n_threads: Option<usize>,
```
tbrezot, line 150: these are not threads anymore:
```suggestion
    n_workers: Option<usize>,
```

src/adt/test_utils.rs
HatemMn: i don't mind but what changes upon this ?

tbrezot (approved):

PR #126, opened 2025-03-04
[4]Feat/postgrestore

HatemMn:
### Batch_Read : 

This is the "simpler" version of batch_read
It is 20% slower but the code is easier

```rust
    async fn batch_read(
        &self,
        addresses: Vec<Self::Address>,
    ) -> Result<Vec<Option<Self::Word>>, Self::Error> {
        let client = self.pool.get().await?;
        // in Postgres databases, statements are cached per connection and not per pool
        let stmt = client
            .prepare_cached(
                format!(
                    "SELECT a,w
                    FROM {0}
                    WHERE a = ANY($1::bytea[])
                    ;",
                    self.table_name
                )
                .as_str(),
            )
            .await?;

        let results = client
            .query(&stmt, &[&addresses
                .iter()
                .map(|addr| addr.as_slice())
                .collect::<Vec<_>>()])
            .await?
            .iter()
            .map(|row| {
                // TODO: REMOVE UNWRAP AND IMPLEMENT THE NECESSARY ERROR HANDLING
                let a = Address::from(
                    <&[u8] as TryInto<[u8; ADDRESS_LENGTH]>>::try_into(row.try_get("a").unwrap())
                        .unwrap(),
                );
                let w: Self::Word = row.try_get::<_, &[u8]>("w").unwrap().try_into().unwrap();
                (a, w)
            })
            .collect::<HashMap<_, _>>();

        Ok(addresses
            .iter()
            // Copying is necessary here since the same word could be returned multiple times.
            .map(|addr| results.get(addr).copied())
            .collect())
    }
```

## Benching batch_read : 

The following test helps run barch_read multiple times with the help of the below companion script

Test : 
```rust
    fn gen_bytes<const BYTES_LENGTH: usize>(rng: &mut impl RngCore) -> [u8; BYTES_LENGTH] {
        let mut bytes = [0; BYTES_LENGTH];
        rng.fill_bytes(&mut bytes);
        bytes
    }
    #[tokio::test]
    async fn bench_batch_read() -> Result<(), PostgresMemoryError> {
        setup_and_run_test("findex_bench_batch_read", |m| async move {
            const NUM_ADDRESSES: usize = 100;
            const NUM_ITERATIONS: usize = 1000;
            let mut rng = StdRng::from_seed([0; KEY_LENGTH]);

            // Generate random addresses
            let addresses: Vec<Address<ADDRESS_LENGTH>> = (0..NUM_ADDRESSES)
                .map(|_| Address::from(gen_bytes::<ADDRESS_LENGTH>(&mut rng)))
                .collect();

            // Pre-write data to all addresses
            for address in &addresses {
                let word = gen_bytes::<WORD_LENGTH>(&mut rng);
                m.guarded_write((address.clone(), None), vec![(address.clone(), word)])
                    .await
                    .unwrap();
            }

            // Benchmark batch reads
            let start = tokio::time::Instant::now();
            for _ in 0..NUM_ITERATIONS {
                m.batch_read(addresses.clone()).await.unwrap();
            }
            let elapsed = start.elapsed();

            println!("BENCHMARK_TIME:{}:", elapsed.as_millis());
        })
        .await
    }
```

Script : 
```bash
#!/bin/bash               
total=0
count=40 # feel free to edit this
for ((i=1; i<=count; i++)); do
    time_ms=$(cargo test bench_batch_read --features postgres-mem --release -- --nocapture 2>&1 | grep 'BENCHMARK_TIME:' | awk -F: '{print $2}')
    if [ -n "$time_ms" ]; then
        total=$((total + time_ms))
        echo "Run $i: $time_ms ms"
    else
        echo "Run $i: Failed to capture time"
    fi
done
if [ $total -ne 0 ]; then
    average=$((total / count))
    echo "Average time over $count runs: $average ms"
else
    echo "No valid measurements were captured"
fi
```

#### Performance : would using the hstore module be better ?
In theory : 🚫  | Test : (not done)

The answer was no, for the followng reasons : 
- B-tree indexes offer is faster lookups vs hstore GIN [src](https://www.postgresql.org/message-id/D83E55F5F4D99B4A9B4C4E259E6227CD014F995C@AUX1EXC02.apac.experian.local)
- Row-level locking enables higher concurrency than document-level

### Performance : would using another index improve perf ?
In theory : 🆗  | Test : 🚫 

No. In theory, HASH index would increase performance however after testing we notice that the tests become around 10% slower (probably due to indexing overhead ?) so it is better to leave default values as they are

### Performance : would using prepared statement perf ?
In theory : 🆗  | Test : 🚀 🆗 

using prepared statements improved performance by around a **huge x5** ! making it comparable with sqlite3 interface

Prepared statements are a good idea since they benefit from caching and remove request processing overhead, which is very useful with our requests that are particularely complex and have a lot of CTE

### Performance : fillfactor ?
In theory : 🚫  | Test : ?

Let's think about what the fillfactor serves : avoiding fragmentation problems when we do frequent update. GWrite is subject to do frequent updates on certain scenarios.
However, the WORD_LENGTH is fixed, meaning no update will ever cause fragmentation issues because the "new data" will always have fixed size ! so in theory, the fillfactor should stay 100% like the default. For reasons that are yet not known, a small perf increase was seen with FILLFACTOR set to 90%

### Performance : storage mode and compression
In theory : 🆗  | Test : Nothing changed

EXTERNAL storage mode stores data out-of-line in a secondary TOAST table but does not compress it
This is pretty cummon tu use with BYTEA values that do not benefit from compression (cpu sees that the compressed data is bigger than the original data and cancels the compression, which is a waste of cpu time - [source and benches](https://www.cybertec-postgresql.com/en/binary-data-performance-in-postgresql/#alternatives-for-storing-binary-data) )

However, changing those parameters didn't make the tests run faster.

PS : the consisty garantees are **not** violated with this storage method

### Alt approach

This is (for reference only) an alternative approach to do G_write

I has, however, poor performance. So similar approaches are to be avoided ( 857.55s, instead of 120 for 1000 workers )

```rust

        let (ag, wg) = (guard.0, guard.1.as_ref().map(|w| &w[..]));

        let params = match &wg {
            Some(guard) => vec![&*ag as &(dyn ToSql + Sync), guard],
            None => vec![&*ag, &None::<&[u8]> as &(dyn ToSql + Sync)],
        }
        .iter()
        .copied()
        .chain(bindings.iter().flat_map(|(a, w)| vec![&**a as _, &*w as _]))
        .collect::<Vec<&(dyn ToSql + Sync)>>();

        let query = &format!(
            "
    WITH
    guard_check AS (
        SELECT w FROM findex_db WHERE a = $1::bytea
    ),
    insert_condition AS (
        INSERT INTO findex_db (a, w)
        SELECT * FROM (
            VALUES
            {}
        ) AS new_values(a, w)
WHERE (
    $2::bytea IS NULL AND NOT EXISTS (SELECT 1 FROM guard_check)
) OR (
    $2::bytea IS NOT NULL AND EXISTS (
       SELECT 1 FROM guard_check WHERE w = $2::bytea
    )
)
        ON CONFLICT (a) DO UPDATE SET w = EXCLUDED.w
    )
    SELECT COALESCE((SELECT w FROM guard_check)) AS original_guard_value;
",
            (3..=params.len())
                .step_by(2) // params.len() is always even
                .map(|i| format!("(${}::bytea, ${}::bytea)", i, i + 1))
                .collect::<Vec<String>>()
                .join(",")
        );
        self.client
            .query_opt(query, &params)
            .await?
            .map_or(Ok(None), |row| {
                row.try_get::<_, Option<&[u8]>>(0)?
                    .map(|b| b.try_into())
                    .map_or(Ok(None), |r| Ok(Some(r?)))
            })
```

### DIFF
--- a/.github/workflows/hack.yml
+++ b/.github/workflows/hack.yml
@@ -7,6 +7,19 @@ jobs:
   hack:
     runs-on: ubuntu-latest
     services:
+      postgres:
+        image: postgres:14
+        env:
+          POSTGRES_USER: cosmian
+          POSTGRES_PASSWORD: cosmian
+          POSTGRES_DB: cosmian
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
       redis:
         image: redis
         options: >-
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 pkg/
+.cargo_check/
 target/
 **/Cargo.lock
 perf*
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ rust-mem = []
 redis-mem = ["redis"]
 sqlite-mem = ["async-sqlite"]
 test-utils = ["tokio", "criterion", "futures", "rand", "rand_distr"]
+postgres-mem = ["tokio-postgres", "tokio/rt-multi-thread", "deadpool-postgres"]
 
 [dependencies]
 aes = "0.8"
@@ -40,11 +41,15 @@ tokio = { version = "1.44", features = ["rt", "rt-multi-thread"], optional = tru
 
 # Memory dependencies
 async-sqlite = { version = "0.5", optional = true }
+deadpool-postgres = { version = "0.14.1", optional = true }
 redis = { version = "0.28", features = [
     "aio",
     "connection-manager",
     "tokio-comp",
 ], optional = true }
+tokio-postgres = { version = "0.7.9", optional = true, features = [
+    "array-impls",
+] }
 
 
 [dev-dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,9 @@ pub use memory::{RedisMemory, RedisMemoryError};
 #[cfg(feature = "sqlite-mem")]
 pub use memory::{SqliteMemory, SqliteMemoryError};
 
+#[cfg(feature = "postgres-mem")]
+pub use memory::{PostgresMemory, PostgresMemoryError};
+
 #[cfg(any(test, feature = "test-utils"))]
 pub use encoding::{
     dummy_encoding::{WORD_LENGTH, dummy_decode, dummy_encode},
--- a/src/memory/mod.rs
+++ b/src/memory/mod.rs
@@ -12,6 +12,11 @@ mod redis_store;
 pub use redis_store::{RedisMemory, RedisMemoryError};
 
 #[cfg(feature = "sqlite-mem")]
-pub mod sqlite_store;
+mod sqlite_store;
 #[cfg(feature = "sqlite-mem")]
 pub use sqlite_store::{SqliteMemory, SqliteMemoryError};
+
+#[cfg(feature = "postgres-mem")]
+mod postgresql_store;
+#[cfg(feature = "postgres-mem")]
+pub use postgresql_store::{PostgresMemory, PostgresMemoryError};
--- a/src/memory/postgresql_store/error.rs
+++ b/src/memory/postgresql_store/error.rs
@@ -1,0 +1,82 @@
+use std::fmt;
+
+#[derive(Debug)]
+pub enum PostgresMemoryError {
+    TokioPostgresError(deadpool_postgres::tokio_postgres::Error),
+    TryFromSliceError(std::array::TryFromSliceError),
+    BuildPoolError(deadpool_postgres::BuildError),
+    GetConnectionFromPoolError(deadpool_postgres::PoolError),
+    PoolConfigError(deadpool_postgres::ConfigError),
+    RetryExhaustedError(usize),
+    InvalidDataLength(usize),
+    TableCreationError(u64),
+}
+
+impl std::error::Error for PostgresMemoryError {}
+
+impl fmt::Display for PostgresMemoryError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::TokioPostgresError(err) => write!(f, "tokio-postgres error: {}", err),
+            Self::TryFromSliceError(err) => write!(f, "try_from_slice error: {}", err),
+            Self::InvalidDataLength(len) => {
+                write!(
+                    f,
+                    "invalid data length: received {} bytes from db instead of WORD_LENGTH bytes",
+                    len
+                )
+            }
+            Self::PoolConfigError(err) => {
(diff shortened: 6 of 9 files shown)


HatemMn:
For reference, stored function for gwrite that can cause ccr bugs :

```sql
         CREATE OR REPLACE FUNCTION guarded_write_with_retry(
                    guard_addr bytea, 
                    guard_value bytea, 
                    addresses bytea[], 
                    valuesAA bytea[]
                ) RETURNS bytea AS $$
                DECLARE
                    retry_count INT := 0;
                    max_retries INT := 10;
                    original_value bytea;
                    backoff_time INT;
                BEGIN
          --      SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
                    WHILE retry_count < max_retries LOOP
                        BEGIN
                            -- Start transaction with appropriate isolation
                            
                            WITH guard_check AS (
                                SELECT w FROM findex_db WHERE a = guard_addr  FOR UPDATE
                            ),
                            dedup_input_table AS (
                                SELECT DISTINCT ON (a) a, w
                                FROM UNNEST(addresses, valuesAA) WITH ORDINALITY AS t(a, w, order_idx)
                                ORDER BY a, order_idx DESC
                            ),
                            insert_cte AS (
                                INSERT INTO findex_db (a, w)
                                SELECT a, w FROM dedup_input_table AS t(a,w)
                                WHERE (
                                    guard_value IS NULL AND NOT EXISTS (SELECT 1 FROM guard_check)
                                ) OR (
                                    guard_value IS NOT NULL AND EXISTS (
                                        SELECT 1 FROM guard_check WHERE w = guard_value
                                    )
                                )
                                ON CONFLICT (a) DO UPDATE SET w = EXCLUDED.w
                            )
                            SELECT COALESCE((SELECT w FROM guard_check)) INTO original_value;
                            
                            -- Exit the retry loop if successful
                            EXIT;
                        EXCEPTION WHEN serialization_failure THEN
                            -- Increment retry counter
                            retry_count := retry_count + 1;
                            
                            -- Calculate exponential backoff with jitter
                            backoff_time := retry_count * 10 + floor(random() * 10)::int;
                            
                            -- Perform server-side sleep
                            PERFORM pg_sleep(backoff_time );
                            
                            -- Continue to next iteration
                            CONTINUE;
                        END;
                    END LOOP;
                    
                    RETURN original_value;
                END;
                $$ LANGUAGE plpgsql;
```

and then : 
```rust
const G_WRITE_STMT_smol: &str = "SELECT guarded_write($1, $2, $3, $4) as original_value";
```

HatemMn (commented):
TODO yet : 
- rebase
- gwrite as db function

tbrezot (approved):

PR #134, opened 2025-04-08
[2]feat: use tostring to easy guard condition

HatemMn:
This PR fixes a very complicated bug that happens when benching the redis interface.

TL;DR : when we flush the DB while it still has pending write operations to perform, the value==false condition can be triggered leaving to an inconsistent state.

I got rid of (value == false) by converting everything to string using built-in **tostring** function available in LUA

Reminder : as a design choice made in the MemoryADT implementation for Redis, we map None guards to a binary false ( see line 126 of redis_store.rs ).

In the previous code, the guard_value == value check was not sufficient because when there is "no guard", it will be comparing nil to b"false", which valuates to false (we don't want that). This is why value==false check was necessary (in LUA nil **always** evaluates to false, so comparing nil to a binary false will yield **true** .

Also, as far as I know, there is no way to pass nil directily via rust

So this method does the trick ( PS : the benches do pass 🥳 The workflow was tested here : https://github.com/Cosmian/findex/actions/runs/14335731155/job/40182250331 )

PS : an issue was opening to redis-rs about this topic : https://github.com/redis-rs/redis-rs/issues/1605

### DIFF
--- a/src/memory/redis_store.rs
+++ b/src/memory/redis_store.rs
@@ -13,7 +13,7 @@ local guard_word    = ARGV[2]
 local length        = ARGV[3]
 local current_word  = redis.call('GET',guard_address)
 
--- If no value is found, nil is converted to 'false'.
+-- If no word is found, nil is converted to 'false'.
 if guard_word == tostring(current_word) then
     for i = 4,(length*2)+3,2
     do


src/memory/redis_store.rs
```diff
@@ -9,21 +9,18 @@ use std::{fmt, marker::PhantomData};
...
+local length        = ARGV[3]
+local current_word  = redis.call('GET',guard_address)
 
-local value = redis.call('GET',guard_address)
-
--- compare the provided guard to the currently stored value
--- if no value is found, nil will be converted to 'false' by the tostring function
-if guard_value == tostring(value) then
-    -- guard passed, loop over bindings and insert them
+-- If none is found, convert nil to 'false'.
```
HatemMn, line 16: I find this sentence confusing, there is nothing called **none** in LUA, if we wanted to verify the non existance of a value in the said address, we pass (exactly) the binary representation of "false" (which is  [102, 97, 108, 115, 101] ) and if compare to the toString(...) of what is found on that address - if nil is found, toString will convert it to a "false", and the condition will pass

If minimalism and non verbove comments is our goal, I suggest keeping `-- if no value is found at guard_address, nil will be converted to 'false' by the tostring function` and deleting the rest

src/adt/test_utils.rs
```diff
@@ -110,13 +110,11 @@ pub async fn test_wrong_guard<const WORD_LENGTH: usize, Memory>(
     let a = Memory::Address::from(gen_bytes(&mut rng));
     let w = Memory::Word::from(gen_bytes(&mut rng));
 
-    // write a value to some address
```
HatemMn: (for info only)
I don't mind taking those out

But are they bad practice ? I remember vaguely that cloudproof js had this sort of comments, which made reading its tests much more easy for me and same here when i need to debug

tbrezot: These duplicate the function doc.

src/memory/redis_store.rs
tbrezot: Changed to `If no value is found, nil is converted to 'false'.` to keep it short.

tbrezot (approved):

PR #135, opened 2025-04-10
[5] Bench/postgresql benches

HatemMn:
## Goal :

Add proper (refactored) postgresql benches

Most recent result : 
https://github.com/Cosmian/findex/actions/runs/14713210309

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -410,7 +410,6 @@ criterion_group!(
     bench_search_multiple_keywords,
     bench_insert_multiple_bindings,
     bench_contention,
-    bench_one_to_many,
 );
 
 criterion_main!(benches);


HatemMn (commented):
Nothing to signal

The related bench is this one : https://github.com/Cosmian/findex/actions/runs/14471064240

As we know, can take some time. Only merger after bench is completed

.github/workflows/hack.yml
tbrezot, line 1: revert this diff

benches/benches.rs
```diff
@@ -14,24 +10,70 @@ use cosmian_findex::InMemory;
...
 #[cfg(feature = "redis-mem")]
 use cosmian_findex::RedisMemory;
+/// Gets the Redis URL to use based on the following priority:
+/// 1. If REDIS_HOST environment variable is set, use "redis://{REDIS_HOST}:6379"
+/// 2. If no environment variable is set but a Redis server is reachable at
+///    localhost (127.0.0.1:6379), use "redis://127.0.0.1:6379"
+/// 3. If neither of the above conditions are met, fall back to the default
+///    dockered Redis service at "redis://redis:6379"
+#[cfg(feature = "redis-mem")]
+fn get_redis_url() -> String {
```
tbrezot, line 25: I prefer the old version since we do not hard-code anything but a default. The CI is to be fixed accordingly. Except if the old version cannot work, I would rather like you to revert this change.

benches/benches.rs
```diff
@@ -14,24 +10,70 @@ use cosmian_findex::InMemory;
...
+}
 
-#[cfg(feature = "sqlite-mem")]
-const SQLITE_PATH: &str = "./target/benches.sqlite";
+/// To run the postgresql benchmarks locally, add the following service to your pg_service.conf file (usually under ~/.pg_service.conf):
+/// [cosmian_service]
+/// host=localhost
+/// dbname=cosmian
+/// user=cosmian
+/// password=cosmian
```
tbrezot, line 45: This doc should be next to the PgSQL memory implementation. Since anyway it is need to pass the tests.

benches/benches.rs
```diff
@@ -14,24 +10,70 @@ use cosmian_findex::InMemory;
...
+///
+/// Gets the PostgreSQL URL to use based on the following priority:
+/// 1. If POSTGRES_HOST environment variable is set, use "postgres://cosmian:cosmian@{POSTGRES_HOST}/cosmian"
+/// 2. If no environment variable is set but a PostgreSQL server is reachable at
+///    localhost (127.0.0.1:5432), use "postgres://cosmian:cosmian@localhost/cosmian"
+/// 3. If neither of the above conditions are met, fall back to the default
+///    dockered PostgreSQL service at "postgres://cosmian:cosmian@postgres/cosmian"
+#[cfg(feature = "postgres-mem")]
```
tbrezot, line 53: If you revert the Redis change, implement this method in a similar way.

benches/benches.rs
```diff
@@ -237,43 +352,105 @@ mod delayed_memory {
...
+    );
+
+    #[cfg(feature = "postgres-mem")]
+    bench_memory_one_to_many(
+        "Postgres",
+        N_PTS,
+        async || {
+            let m = PostgresMemory::connect_and_init_table(
+                get_postgresql_url(),
+                "bench_memory_contention_m_1_var_1".to_string(),
```
tbrezot, line 410: Maybe no need for such a name, it is really hard to parse. Calling them all the same name should work. Plus it is misleading since there is a bench called `bench_contention` but this is not the one.

benches/benches.rs
```diff
@@ -237,43 +352,105 @@ mod delayed_memory {
...
+        "Postgres",
+        N_PTS,
+        async || {
+            let m = PostgresMemory::connect_and_init_table(
+                get_postgresql_url(),
+                "bench_memory_contention_m_10_var_5".to_string(),
+            )
+            .await
+            .unwrap();
+            DelayedMemory::new(m, 10, 5)
```
tbrezot, line 450: We could consider factorizing all these instantiations in a loop over a set of delay parameters. This will put forward the fact that these delays are the same for all memories and will save many lines.

src/memory/postgresql_store/memory.rs
```diff
@@ -74,6 +74,49 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize>
             _marker: PhantomData,
         })
     }
+
+    // Initialize the table in the database, used only for benching
```
tbrezot, line 78: Then it should be in the bench file.

src/memory/postgresql_store/memory.rs
```diff
@@ -74,6 +74,49 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize>
...
+        let test_pool = pg_config.builder(NoTls)?.build()?;
+
+        let m = Self::connect_with_pool(test_pool, table_name.to_string()).await?;
+
+        m.initialize_table(db_url.to_string(), table_name, NoTls)
+            .await?;
+        Ok(m)
+    }
+
+    /// Deletes all rows from a given table, used only for benches
```
tbrezot, line 98: same

src/memory/sqlite_store.rs
tbrezot, line 1: No, for the second time (I corrected it myself in your other MR) !

src/memory/postgresql_store/memory.rs
```diff
@@ -74,6 +74,49 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize>
...
+                "CREATE TABLE IF NOT EXISTS {} (
+                    a BYTEA PRIMARY KEY CHECK (octet_length(a) = {}),
+                    w BYTEA NOT NULL CHECK (octet_length(w) = {})
+                );",
+                self.table_name, ADDRESS_LENGTH, WORD_LENGTH
+            ),
+            &[],
+        )
+        .await?;
+        Ok(())
```
tbrezot, line 118: No, I am almost sure there is a way in SQL for atomically dropping everything in a table.

src/memory/encryption_layer.rs
tbrezot, line 1: This change is okay, but for the next time, it is always best to leave untouched files you do not modify.

benches/benches.rs
```diff
@@ -237,43 +352,105 @@ mod delayed_memory {
...
-    #[cfg(feature = "redis-mem")]
+    #[cfg(any(feature = "redis-mem", feature = "postgres-mem"))]
     use delayed_memory::*;
 
     #[cfg(feature = "redis-mem")]
     bench_memory_one_to_many(
         "Redis",
         N_PTS,
-        async || DelayedMemory::new(RedisMemory::connect(&url).await.unwrap(), 1, 1),
+        async || DelayedMemory::new(RedisMemory::connect(&get_redis_url()).await.unwrap(), 1, 1),
```
tbrezot, line 377: I personally preferred the old version since it fits in half my screen... but okay.

.github/workflows/benches.yml
```diff
@@ -1,5 +1,6 @@
 ---
 name: Benches
+run-name: Benchmark on ${{ github.ref_name }} - ${{ github.sha }}
```
tbrezot, line 3: What is this used for?

src/test_utils/benches.rs
```diff
@@ -328,14 +328,11 @@ pub fn bench_memory_one_to_many<
...
+    const MAX_VAL: usize = 100;
     // Redis memory backend requires a tokio runtime, and all operations to
     // happen in the same runtime, otherwise the connection returns a broken
     // pipe error.
-    let rt = Builder::new_multi_thread()
-        .worker_threads(8)
-        .enable_all()
-        .build()
-        .unwrap();
+    let rt = Builder::new_multi_thread().enable_time().build().unwrap();
```
HatemMn, line 335: solve this

.github/workflows/benches.yml
HatemMn: This gives the benchmark a distinct name for each run instead of many benches with the same exact name

src/memory/encryption_layer.rs
HatemMn: okay it was just a matter of form

HatemMn (commented):
I just ran a bench for the post review : https://github.com/Cosmian/findex/actions/runs/14497790270

It should be over by tmrw morning

Cargo.toml
```diff
@@ -25,24 +25,28 @@ rust-mem = []
...
 rand_distr = { version = "0.5.1", optional = true }
-tokio = { version = "1.44", features = ["rt", "rt-multi-thread"], optional = true }
+tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
+
 
 # Memory dependencies
 async-sqlite = { version = "0.5", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
-redis = { version = "0.28", features = [
+redis = { version = "0.28.2", features = [
```
tbrezot, line 49:
```suggestion
redis = { version = "0.28", features = [
```

Cargo.toml
```diff
@@ -53,13 +57,8 @@ tokio-postgres = { version = "0.7.9", optional = true, features = [
...
 [dev-dependencies]
-futures = "0.3"
-tokio = { version = "1.44", features = [
-    "macros",
-    "rt",
-    "rt-multi-thread",
-    "time",
-] }
+futures = { version = "0.3" }
+tokio = { version = "1.44.0", features = ["macros", "rt-multi-thread"] }
```
tbrezot, line 61:
```suggestion
tokio = { version = "1.44", features = ["macros", "rt-multi-thread"] }
```

benches/benches.rs
tbrezot: My point was that there should be no need to hard-code `redis://redis:6379`. Plus your new function has too much intelligence : since it attempts connecting to `localhost` first. Previous version was simpler: return the URL in the environment variable, defaulting to `localhost`. But it was not less generic since any address could passed through this environment variable.

In short, imho, you made a function more complex in order to fix something that should not be fixed here.

benches/benches.rs
```diff
@@ -237,45 +350,63 @@ mod delayed_memory {
...
-    bench_memory_one_to_many(
-        "Redis",
-        N_PTS,
-        async || DelayedMemory::new(RedisMemory::connect(&url).await.unwrap(), 10, 5),
-        c,
-        DelayedMemory::clear,
-        &mut rng,
-    );
+    for (mean, variance) in &delay_params {
+        #[cfg(feature = "postgres-mem")]
```
tbrezot, line 392:
```suggestion
    #[cfg(feature = "postgres-mem")]
    for (mean, variance) in &delay_params {
```

benches/benches.rs
```diff
@@ -237,45 +350,63 @@ mod delayed_memory {
...
-        N_PTS,
-        async || DelayedMemory::new(RedisMemory::connect(&url).await.unwrap(), 10, 1),
-        c,
-        DelayedMemory::clear,
-        &mut rng,
-    );
+    let delay_params = vec![(1, 1), (10, 1), (10, 5)]; // tuples of (mean, variance)
+
+    for (mean, variance) in &delay_params {
+        #[cfg(feature = "redis-mem")]
```
tbrezot, line 374:
```suggestion
    #[cfg(feature = "postgres-mem")]
    for (mean, variance) in &delay_params {
```

src/memory/postgresql_store/memory.rs
tbrezot: My point was: either your function is useful and generic enough to be used independently from the benches and there is no need to add "only used in bench", either your function is not and it should be moved. I consider `clear` to be simple, generic, and usable in tests etc. I let you apply whichever solution you prefer.

src/memory/postgresql_store/memory.rs
```diff
@@ -217,16 +247,24 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize> MemoryADT
...
 }
-
+/// To run the postgresql benchmarks locally, add the following service to your pg_service.conf file (usually under ~/.pg_service.conf):
+/// [cosmian_service]
+/// host=localhost
+/// dbname=cosmian
+/// user=cosmian
+/// password=cosmian
 #[cfg(test)]
 mod tests {
```
tbrezot, line 257: Module docs is usually written like that (commit this change and run `cargo fmt`):

```suggestion
#[cfg(test)]
mod tests {
//! To run the postgresql benchmarks locally, add the following service to your pg_service.conf file
//! (usually under ~/.pg_service.conf):
//!
//! [cosmian_service]
//! host=localhost
//! dbname=cosmian
//! user=cosmian
//! password=cosmian
```

src/memory/postgresql_store/memory.rs
HatemMn: Okay now i understand better. After reconsideration I think the "clear" function is very useful for tests but the other one can only used in benches and is redundant so I moved it to the bench file

benches/benches.rs
```diff
@@ -287,6 +410,7 @@ criterion_group!(
     bench_search_multiple_keywords,
     bench_insert_multiple_bindings,
     bench_contention,
+    bench_one_to_many,
```
tbrezot, line 413: Why is this benched a second time ?

tbrezot (commented):
All good for me, except for a possibly duplicated bench.

benches/benches.rs
HatemMn: good catch ! done

PR #137, opened 2025-04-14
fix: downgrade async-sqlite to 4.0.0

HatemMn:
don't merge this

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,8 @@ tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 
 
 # Memory dependencies
+# Due to dependency tree issues with sqlx requiring a lower version of `libsqlite3-sys`, do not upgrade the `async-sqlite` version above 0.4
+# before checking the update will not break findex-server's build.
 async-sqlite = { version = "0.4", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
 redis = { version = "0.28", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,9 +44,9 @@ tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 
 
 # Memory dependencies
-async-sqlite = { version = "0.5", optional = true }
+async-sqlite = { version = "0.4.0", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
-redis = { version = "0.28", features = [
+redis = { version = "0.28.2", features = [
     "aio",
     "connection-manager",
     "tokio-comp",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 # Memory dependencies
 async-sqlite = { version = "0.4", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
-redis = { version = "0.28.2", features = [
+redis = { version = "0.28", features = [
     "aio",
     "connection-manager",
     "tokio-comp",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 
 
 # Memory dependencies
-async-sqlite = { version = "0.4.0", optional = true }
+async-sqlite = { version = "0.4", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
 redis = { version = "0.28.2", features = [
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -44,9 +44,9 @@ tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 
 
 # Memory dependencies
-async-sqlite = { version = "0.5", optional = true }
+async-sqlite = { version = "0.4", optional = true }
 deadpool-postgres = { version = "0.14.1", optional = true }
-redis = { version = "0.28", features = [
+redis = { version = "0.28.2", features = [
```
tbrezot, line 49:
```suggestion
redis = { version = "0.28", features = [
```

src/memory/sqlite_store.rs
```diff
@@ -48,42 +49,70 @@ impl<Address, Word> Debug for SqliteMemory<Address, Word> {
...
+        pool.conn(move |conn| conn.execute_batch(&create_table_script(FINDEX_TABLE_NAME)))
             .await?;
 
         Ok(Self {
             pool,
             _marker: PhantomData,
         })
     }
+
+    /// Creates the memory's table using a connection from the given pool.
```
tbrezot, line 92: I don't feel the current doc describes what the functions does. Maybe:
```suggestion
    /// Returns an `SqliteMemory` instance storing data in a table with the given name
    /// and connecting to the DB using connections from the given pool.
```

src/memory/sqlite_store.rs
```diff
@@ -145,13 +174,15 @@ impl<const ADDRESS_LENGTH: usize, const WORD_LENGTH: usize> MemoryADT
                 )?;
 
                 let current_word = tx
-                    .query_row("SELECT w FROM memory WHERE a = ?", [&*ag], |row| row.get(0))
+                    .query_row("SELECT w FROM findex_memory WHERE a = ?", [&*ag], |row| {
```
tbrezot, line 177: Why aren't you using your constant ?

tbrezot:
I think you should add a comment on the line declaring the SQLite dependency to explain it should not be upgraded for compatibility reasons.

I am still reluctant to accept this downgrade since it imports a constraint from another repository, breaking the usual flow. Said otherwise, it makes no sense for Findex to be constrained by a repository it doesn't (and shouldn't) know about. Did you consider relocating the SQLite memory in your repo? The design goal of having a clean memory interface was to allow users to implement memories that would best fit their needs without requiring upstream modifications or fork.

HatemMn (commented):
> I am still reluctant to accept this downgrade since it imports a constraint from another repository, breaking the usual flow. Said otherwise, it makes no sense for Findex to be constrained by a repository it doesn't (and shouldn't) know about. Did you consider relocating the SQLite memory in your repo? The design goal of having a clean memory interface was to allow users to implement memories that would best fit their needs without requiring upstream modifications or fork.

I agree, what you said makes sense, and there are many solutions but the best solution is what will align better with our mid to long term vision on the `findex` and `findex-server` projects. We can also create a dedicated branch, to be rebased on `main` regulary to avoid forking.

Relocating the memory will cause some code duplication, but will leave me more freedom to meddle with low level Pragmas. Anyway, I think this a metter to discuss with @Manuthor too

src/memory/sqlite_store.rs
```diff
@@ -48,34 +51,63 @@ impl<Address, Word> Debug for SqliteMemory<Address, Word> {
...
+
+    /// Creates the memory's table using a connection from the given pool.
+    ///
+    /// # Arguments
+    ///
+    /// * `pool` - The pool to use for the memory.
+    /// * `table_name` - The name of the table to create.
+    pub async fn connect_with_pool(
+        pool: Pool,
+        table_name: String,
```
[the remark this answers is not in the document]
tbrezot: I was talking about the string

Manuthor:
> > I am still reluctant to accept this downgrade since it imports a constraint from another repository, breaking the usual flow. Said otherwise, it makes no sense for Findex to be constrained by a repository it doesn't (and shouldn't) know about. Did you consider relocating the SQLite memory in your repo? The design goal of having a clean memory interface was to allow users to implement memories that would best fit their needs without requiring upstream modifications or fork.
> 
> I agree, what you said makes sense, and there are many solutions but the best solution is what will align better with our mid to long term vision on the `findex` and `findex-server` projects. We can also create a dedicated branch, to be rebased on `main` regulary to avoid forking.
> 
> Relocating the memory will cause some code duplication, but will leave me more freedom to meddle with low level Pragmas. Anyway, I think this a metter to discuss with @Manuthor too

What about creating a new repo for Findex memories?

PR #138, opened 2025-05-26
Refactor/migrate findex memories

HatemMn:
Related : https://github.com/Cosmian/findex-memories/pull/1

PR #139, opened 2025-05-27
chore: less dependencies

HatemMn:
This is a preparatory pr for : https://github.com/Cosmian/findex/pull/138

It was cut aside because it is subject to debate and we can omit it if there is a valid motivation to keep **futures**

## Motivation : 
- Reduce dependencies by deleting futures crate from dependency list without impacting any functionnality
- Do exactly what the futures crate used to do without adding more deps

## Reasons why : 
-[ Tokio re-exports what it needs from the futures crate](https://users.rust-lang.org/t/relationship-between-std-futures-futures-and-tokio/38077), so (in general) we don't need both
- **block_on** blocks the whole thread and prevents the runtime from optimising the ressources
- Long quest of having the less dependencies for the findex repository

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,8 @@ path = "src/lib.rs"
 rust-mem = []
 redis-mem = ["redis"]
 sqlite-mem = ["async-sqlite"]
-test-utils = ["tokio", "criterion", "rand", "rand_distr"]
-postgres-mem = ["tokio-postgres", "tokio", "deadpool-postgres"]
+test-utils = ["criterion", "rand", "rand_distr", "agnostic/tokio"]
+postgres-mem = ["tokio-postgres", "deadpool-postgres"]
 
 [dependencies]
 aes = "0.8"
@@ -39,7 +39,6 @@ xts-mode = "0.5"
 criterion = { version = "0.6", optional = true }
 rand = { version = "0.9.0", optional = true }
 rand_distr = { version = "0.5.1", optional = true }
-tokio = { version = "1.44", features = ["rt-multi-thread"], optional = true }
 
 # Memory dependencies
 async-sqlite = { version = "0.5", optional = true }
@@ -52,9 +51,11 @@ redis = { version = "0.28", features = [
 tokio-postgres = { version = "0.7.9", features = [
     "array-impls",
 ], optional = true }
+agnostic = { version = "0.7", optional = true, features = ["tokio", "smol"] }
 
 
 [dev-dependencies]
+agnostic = { version = "0.7", features = ["tokio", "smol"] }
 tokio = { version = "1.44", features = ["macros", "rt-multi-thread"] }
 
 [[bench]]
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -35,13 +35,10 @@ cosmian_crypto_core = { version = "10.1", default-features = false, features = [
 ] }
 xts-mode = "0.5"
 
-# Used in benches and tests.
-criterion = { version = "0.5", optional = true }
-futures = { version = "0.3", optional = true }
+# Used only in benches and tests
```
tbrezot, line 38: Revert this change
```suggestion
# Used in benches and tests.
```

src/test_utils/benches.rs
```diff
@@ -139,8 +141,8 @@ pub fn bench_memory_search_multiple_keywords<
                             let findex = findex.clone();
                             handles.push(tokio::spawn(async move { findex.search(&kw).await }))
                         }
-                        for res in join_all(handles).await {
-                            res.unwrap().unwrap();
+                        for res in handles {
+                            res.await.expect("Search task failed").unwrap();
```
tbrezot, line 145: No, this first expect unwraps an error linked to join failure.

src/test_utils/benches.rs
```diff
@@ -256,10 +258,14 @@ pub fn bench_memory_contention<
...
-                                tokio::spawn(async move { findex.insert(kw, vs).await })
-                            }))
-                            .await
+                            let handles = iterator
+                                .map(|(findex, (kw, vs))| {
+                                    tokio::spawn(async move { findex.insert(kw, vs).await })
+                                })
+                                .collect::<Vec<_>>();
+                            for h in handles {
+                                h.await.expect("Insert task failed").unwrap();
```
tbrezot, line 267: same

tbrezot: check everywhere

src/test_utils/memory_tests.rs
```diff
@@ -230,16 +256,18 @@ pub async fn test_rw_same_address<const WORD_LENGTH: usize, Memory>(
 /// * `memory` - The Memory ADT implementation to test.
 /// * `seed` - The seed used to initialize the random number generator.
 /// * `n_threads` - The number of threads to spawn. If None, defaults to 100.
-pub async fn test_guarded_write_concurrent<const WORD_LENGTH: usize, Memory>(
+pub async fn test_guarded_write_concurrent<const WORD_LENGTH: usize, Memory, S>(
     memory: &Memory,
     seed: [u8; SEED_LENGTH],
     n_threads: Option<usize>,
+    spawer: &S,
```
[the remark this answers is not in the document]
tbrezot: You can keep the type as interface. However, since instances of your type contains no information, you actually don't need to pass it along.

src/test_utils/benches.rs
HatemMn: I will rectify this everywhre

HatemMn: done

HatemMn:
It's complicated to continue working here since the whole direction of this pr has changed

Moving the new work into this one : https://github.com/Cosmian/findex/pull/141

PR #142, opened 2025-06-13
Refacto before publication

tbrezot:
- rationalize dependencies (remove redundant features, move dependencies in dev where possible, group dependencies by features...);
- upgrade toolchain (now using v1.87);
- upgrade all dependencies;
- refactor PgSQL memory and others (minors)

# Breaking change

Both SQL memories now need to be *initialized* before use.

### DIFF
--- a/crate/memories/Cargo.toml
+++ b/crate/memories/Cargo.toml
@@ -17,33 +17,32 @@ path = "src/lib.rs"
 [features]
 redis-mem = ["redis"]
 sqlite-mem = ["async-sqlite"]
-postgres-mem = ["tokio-postgres", "tokio", "deadpool-postgres"]
-test-utils = ["cosmian_findex/test-utils"]
+postgres-mem = ["deadpool-postgres", "tokio", "tokio-postgres"]
+test-utils = ["cosmian_findex/test-utils", "redis/tokio-comp"]
 
 [dependencies]
-cosmian_crypto_core.workspace = true
+async-sqlite = { version = "0.5", optional = true }
 cosmian_findex = { path = "../findex", version = "8.0" }
-async-sqlite = { version = "0.4", optional = true } # `async-sqlite` dependency is pinned at v0.4 due to dependency tree issues with sqlx requiring a lower version of `libsqlite3-sys`
-deadpool-postgres = { version = "0.14.1", optional = true }
+deadpool-postgres = { version = "0.14", optional = true }
 redis = { version = "0.32", default-features = false, features = [
-    "aio",
     "connection-manager",
     "tokio-comp",
 ], optional = true }
 tokio = { workspace = true, features = ["rt-multi-thread"], optional = true }
-tokio-postgres = { version = "0.7.9", optional = true, features = [
+tokio-postgres = { version = "0.7", optional = true, features = [
     "array-impls",
 ] }
 
 [dev-dependencies]
+cosmian_crypto_core.workspace = true
 cosmian_findex = { path = "../findex", version = "8.0", features = [
     "test-utils",
 ] }
-futures = { version = "0.3" }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 criterion = { workspace = true }
-rand = { version = "0.9.0" }
-rand_distr = { version = "0.5.1" }
+futures = { version = "0.3" }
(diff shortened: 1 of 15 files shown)


crate/memories/src/sqlite_mem.rs
```diff
@@ -47,43 +47,16 @@ impl<Address, Word> Debug for SqliteMemory<Address, Word> {
...
-    /// Builds a pool connected to a known DB (using the given path) and creates
-    /// the `findex_memory` table.
-    ///
-    /// # Arguments
-    ///
-    /// * `path` - The path to the sqlite3 database file.
-    pub async fn connect<P: AsRef<std::path::Path>>(
-        path: P,
+    /// Returns a new memory instance using a pool of connections to the SQLite
+    /// database at the given path.
```
HatemMn, line 52:
```suggestion
    /// Returns a new memory instance with a new pool of connections to the SQLite
    /// database at the given path.
```

crate/memories/src/sqlite_mem.rs
```diff
@@ -92,34 +65,45 @@ impl<Address, Word> SqliteMemory<Address, Word> {
...
+                 a BLOB PRIMARY KEY,
+                 w BLOB NOT NULL
+             );",
+            self.table_name
+        );
+
+        self.pool
+            .conn(move |conn| conn.execute_batch(&table_name))
+            .await?;
+        Ok(())
```
HatemMn, line 98:
```suggestion
        let initialization_script = format!(
            "PRAGMA synchronous = NORMAL;
             PRAGMA journal_mode = WAL;
             CREATE TABLE IF NOT EXISTS {} (
                 a BLOB PRIMARY KEY,
                 w BLOB NOT NULL
             );",
            self.table_name
        );

        self.pool
            .conn(move |conn| conn.execute_batch(&initialization_script))
            .await?;
        Ok(())
```
or "table_creation_script", or something of the sort

crate/memories/src/sqlite_mem.rs
```diff
@@ -92,34 +65,45 @@ impl<Address, Word> SqliteMemory<Address, Word> {
...
             pool,
             table_name,
             _marker: PhantomData,
-        })
+        }
     }
-}
 
-impl<Address: Send + Sync, Word: Send + Sync> SqliteMemory<Address, Word> {
+    /// Makes sure the correct table exist in the associated database.
```
HatemMn, line 78: This comment misled me because by reading it without seeing the signature I would say that the goal is returning a boolean equals "table exists ?"

I would rather suggest smth like :

```suggestion
    /// Creates the memory's table in the associated database if it does not exist.
```

crate/memories/src/postgresql_mem/memory.rs
```diff
@@ -17,74 +13,77 @@ pub struct PostgresMemory<Address, Word> {
...
+                $2::bytea IS NOT NULL AND EXISTS (
+                    SELECT 1 FROM guard_check WHERE w = $2::bytea
+                )
+            )
+            ON CONFLICT (a) DO UPDATE SET w = EXCLUDED.w
+        )
+        SELECT COALESCE((SELECT w FROM guard_check)) AS original_guard_value;",
+            self.table_name
+        )
+    }
```
HatemMn, line 86: This function is used once below, why refactor it ?
image

If the linter is complaining about cognitive complexity, I'm fine with disabling the lint

HatemMn (commented):
All good, almost all comments are about very minor changes regarding comments / style

crate/memories/src/sqlite_mem.rs
tbrezot: Felt okay for me but I will change.

crate/memories/Cargo.toml
```diff
@@ -17,33 +17,38 @@ path = "src/lib.rs"
...
+tokio-postgres = { version = "0.7", optional = true, features = [
     "array-impls",
 ] }
 
+# Needed for Redis memory.
+redis = { version = "0.32", default-features = false, features = [
+    "connection-manager",
+    "tokio-comp"
+], optional = true }
+
```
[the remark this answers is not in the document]
tbrezot: I disagree with your revert: it is best to keep the dependencies sorted by feature imho.

PR #153, opened 2025-08-08
Batch work

HatemMn:
## Updated summary (22/08/2025)

This new implementation of **batcher_findex** keeps the core idea of the previous one, but separates the different "inside" operations following the ideas stated in the PDF below :

[memory_batcher.notes (2).pdf](https://github.com/user-attachments/files/21965040/memory_batcher.notes.2.pdf)

However, due to the Rust compiler strict constraint on **enum**s and the need to have a readable code that distinguishes an input/an output/an operation, some modifications have been done. So please take note of the comments and highlights that were added on top of the original file.

### Other changes : 
- merged the insert and delete tests
- separated buffer errors from BatcherMemory errors
- basically everything written in **operations.rs**

### Next steps : 

- [ ] Drop the **futures** library and go toward a runtime free implementation based on spawning (like findex does)
- [ ] No mention of findex in the batcher --> must be decorrelated
- [ ] Use actual database interfaces, Redis in particular
- [ ] Bench the new algorithm
- [ ] Make a non trivial integration test - As a primary idea : use the [shuttle](https://github.com/awslabs/shuttle) scheduler to test extreme edge cases and assert the code is free of race conditions
- [ ] (optional) provide a correctness proof for this algorithm
- [ ]  (optional)  document


crate/findex/src/memory_layers/batching_layer/batching_layer.rs
```diff
@@ -0,0 +1,304 @@
...
+                    // conflicting implementation in crate `core`
+                    .map_err(BatchingLayerError::<M>::Memory)?;
+
+                // Distribute results to each batch's sender
+                for (input, sender) in batches.into_iter().rev() {
+                    let split_point = all_results.len() - input.len();
+                    // After this call, all_results will be left containing the elements [0, split_point)
+                    // that's why we need to reverse the batches
+                    let batch_results = all_results.split_off(split_point);
+                    sender.send(Ok(batch_results)).map_err(|_| {
```
HatemMn: Used std::mem::take to empty the vector and move ownership of its content at the same time

*this comment is only for clarification. Close it if fine*

HatemMn:
```rust
    pub fn send(self, t: T) -> Result<(), T> {
        self.inner.send(t)
    }
```

the send function has an indeed curious way of using the Err variant

_this comment is only for clarification. Close it if fine_

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,402 @@
...
+            let memory_arc = buffered_memory.clone();
+            let future = async move {
+                // Create a temporary Findex instance using the shared batching layer
+                let findex: Findex<WORD_LENGTH, Value, EncodingError, BatcherArc<BatcherMemory>> =
+                    Findex::<WORD_LENGTH, Value, EncodingError, _>::new(
+                        // this (cheap) Arc cline is necessary because `decrement_capacity` is called
+                        // below and needs to be able to access the Arc
+                        memory_arc.clone(),
+                        *self.encode,
+                        *self.decode,
```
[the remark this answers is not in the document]
HatemMn: this remark is interesting but I am fine with current approach, maybe @tbrezot do you have an opinion about it ?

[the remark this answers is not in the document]
HatemMn: same

crate/findex/README.md
```diff
@@ -2,6 +2,8 @@
 
 This crate provides the core functionality of Findex, defining the abstract data types, cryptographic operations, and encoding algorithms.
 
+Findex also supports batch operations, allowing to index and search multiple items in a single request. This feature improves performance and efficiency when network time is a bottleneck.
```
tbrezot: This is not the reason as `async` should solve that. The reason is that too many file descriptor cannot be open in many linux distributions by default. Also, there might be an over due to instantiating a connection per request or limiting the number of truly parallel requests perform due to the use of a fixed-sized pool.

crate/findex/examples/databases.rs
```diff
@@ -3,7 +3,7 @@
 mod shared_utils;
 
 use cosmian_crypto_core::{CsRng, Secret, reexport::rand_core::SeedableRng};
-use cosmian_findex::{Findex, IndexADT, MemoryEncryptionLayer};
+use cosmian_findex::{Findex, IndexADT, encryption_layer::MemoryEncryptionLayer};
```
tbrezot: why this change?

crate/findex/examples/in_memory.rs
```diff
@@ -3,7 +3,7 @@ use cosmian_crypto_core::{
     CsRng, Secret,
     reexport::rand_core::{CryptoRngCore, SeedableRng},
 };
-use cosmian_findex::{Findex, IndexADT, MemoryEncryptionLayer, Op};
+use cosmian_findex::{Findex, IndexADT, Op, encryption_layer::MemoryEncryptionLayer};
```
tbrezot: why this change?

crate/findex/src/adt.rs
```diff
@@ -1,11 +1,29 @@
...
-//! - the index ADT;
 //! - the vector ADT;
+//! - the index ADT;
+//! - the batched index ADT;
 //!
 //! Each of them strive for simplicity and consistency with the classical CS notions.
 
 use std::{collections::HashSet, future::Future, hash::Hash};
 
+pub trait VectorADT: Send {
```
tbrezot: please do not move code around: revert this change.

crate/findex/src/adt.rs
```diff
@@ -32,21 +50,31 @@ pub trait IndexADT<Keyword: Send + Hash, Value: Send + Hash> {
     ) -> impl Send + Future<Output = Result<(), Self::Error>>;
 }
 
-pub trait VectorADT: Send {
-    /// Vectors are homogeneous.
-    type Value: Send;
+/// BatcherSSEADT (Batched Searchable Symmetric Encryption Abstract Data Type)
```
tbrezot: Unreadable name, and I am not sure this batcher is generic to any SSE call. My guess is that it isn't. So why bother with `SSE`?

crate/findex/src/adt.rs
```diff
@@ -32,21 +50,31 @@ pub trait IndexADT<Keyword: Send + Hash, Value: Send + Hash> {
...
+///
+/// This trait defines batch operations for encrypted searchable indices. It extends
```
tbrezot: what is an encrypted searchable index?

crate/findex/src/adt.rs
tbrezot: I think we settled on something like `IndexBatcher` by opposition to `MemoryBatcher`.

crate/findex/src/adt.rs
```diff
@@ -32,21 +50,31 @@ pub trait IndexADT<Keyword: Send + Hash, Value: Send + Hash> {
...
+/// multiple keywords or entries simultaneously to cut network's overhead and improve performance.
+pub trait BatchedIndexADT<Keyword, Value> {
+    type Error: std::error::Error;
 
-    /// Vector error.
-    type Error: Send + std::error::Error;
+    /// Search the index for the values bound to the given keywords.
+    fn batch_search(
+        &self,
+        keywords: Vec<&Keyword>, // n  --> n fois barch read --> n+1
```
tbrezot: Please read your diffs! This comment is ugly and does not mean anything!
```suggestion
        keywords: Vec<&Keyword>,
```

crate/findex/src/adt.rs
```diff
@@ -32,21 +50,31 @@ pub trait IndexADT<Keyword: Send + Hash, Value: Send + Hash> {
...
+    ) -> impl Future<Output = Result<Vec<HashSet<Value>>, Self::Error>>;
 
-    /// Pushes the given values at the end of this vector.
-    fn push(
-        &mut self,
-        values: Vec<Self::Value>,
+    /// Adds the given values to the index.
```
tbrezot: You need give a more detailed description, something like `Binds each value to their associated keyword in this index`.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
+use crate::adt::BatchedIndexADT;
+use crate::error::BatchFindexError;
+use crate::memory_layers::batching_layer::{BatcherArc, MemoryBatcher};
+use crate::{Decoder, Encoder, Findex, IndexADT};
+use cosmian_sse_memories::{ADDRESS_LENGTH, Address, BatchingMemoryADT};
+use std::sync::atomic::AtomicUsize;
+use std::{collections::HashSet, fmt::Debug, hash::Hash, sync::Arc};
+
+#[derive(Debug)]
+pub struct BatcherFindex<
```
tbrezot: I have already said it but you didn't take it into account: reverse the word order here as this does not respect the usual order in English.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        ));
+
+        for (guard_keyword, bindings) in entries {
+            let memory_arc = buffered_memory.clone();
+            let future = async move {
+                // Create a temporary Findex instance using the shared batching layer
+                let findex: Findex<WORD_LENGTH, Value, EncodingError, BatcherArc<BatcherMemory>> =
+                    Findex::<WORD_LENGTH, Value, EncodingError, _>::new(
+                        // this (cheap) Arc cline is necessary because `decrement_capacity` is called
+                        // below and needs to be able to access the Arc
```
tbrezot: Please take the habit to use a spelling linter on all your comment since it seems you don't bother proof-reading them...
```suggestion
                        // this (cheap) Arc clone is necessary because `decrement_capacity` is called
                        // below and needs to be able to access the Arc
```

tbrezot: Plus all code comments should be complete sentences (thus starting with an upper case and finishing with a dot) or only lower-case words, check the rest of the repo to select the consistent option. And at least be consistent in your own code!

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+                let result = if is_insert {
+                    findex.insert(guard_keyword, bindings).await
+                } else {
+                    findex.delete(guard_keyword, bindings).await
+                };
+
+                // Convert Findex error to BatchingLayerError manually if needed
+                if let Err(findex_err) = result {
+                    return Err(BatchFindexError::Findex(findex_err));
+                }
```
tbrezot: Why do you need manual conversion? Plus your comment repeats your code which is a red flag. Comments should explain the code not describe it.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        let buffered_memory = BatcherArc::new(MemoryBatcher::new_writer(
+            self.memory.clone(),
+            AtomicUsize::new(n),
+        ));
+
+        for (guard_keyword, bindings) in entries {
+            let memory_arc = buffered_memory.clone();
+            let future = async move {
+                // Create a temporary Findex instance using the shared batching layer
+                let findex: Findex<WORD_LENGTH, Value, EncodingError, BatcherArc<BatcherMemory>> =
```
tbrezot: Can't you instantiate a single Findex instance and use it in each task? This would increase readability.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+    async fn batch_insert_or_delete<Keyword>(
+        &self,
+        entries: Vec<(Keyword, impl Send + IntoIterator<Item = Value>)>,
+        is_insert: bool,
+    ) -> Result<(), BatchFindexError<BatcherMemory>>
+    where
+        Keyword: Send + Sync + Hash + Eq,
+    {
+        let mut search_futures = Vec::new();
+        let n = entries.len();
```
tbrezot: This name is just noise.
```suggestion
```

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        let buffered_memory = BatcherArc::new(MemoryBatcher::new_writer(
+            self.memory.clone(),
+            AtomicUsize::new(n),
+        ));
+
+        for (guard_keyword, bindings) in entries {
+            let memory_arc = buffered_memory.clone();
```
tbrezot: This is bad practice to encode types in tames. Use shadowing instead. Plus what is a buffered memory? Isn't it an implementation detail?
```suggestion
            let batch_memory = batch_memory.clone();
```

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+            let future = async move {
+                // Create a temporary Findex instance using the shared batching layer
```
tbrezot: what is a batching layer?

crate/findex/src/batcher_findex.rs
tbrezot: Plus, your comment is wrong: type mismatch.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+                // Convert Findex error to BatchingLayerError manually if needed
+                if let Err(findex_err) = result {
+                    return Err(BatchFindexError::Findex(findex_err));
+                }
+                // once one of the operations succeeds, we should make the buffer smaller
+                memory_arc.decrement_capacity().await?;
+                Ok(())
+            };
+
+            search_futures.push(future);
```
tbrezot: This procedure does not perform any search...

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        // Execute all futures concurrently and collect results
+        futures::future::try_join_all(search_futures).await?;
```
tbrezot: Why not spawning tasks instead? This is what is done in the runtime-generic tests and you are the one who made the move from `join_all` to task spawning.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+    BatcherMemory: Debug
+        + Send
+        + Sync
+        + Clone
+        + BatchingMemoryADT<Address = Address<ADDRESS_LENGTH>, Word = [u8; WORD_LENGTH]>,
+> BatchedIndexADT<Keyword, Value>
+    for BatcherFindex<WORD_LENGTH, Value, EncodingError, BatcherMemory>
+{
+    // type Findex = Findex<WORD_LENGTH, Value, EncodingError, BatcherMemory>;
+    // type BatcherMemory = BatcherMemory;
```
tbrezot: dead code

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        &self,
+        keywords: Vec<&Keyword>,
+    ) -> Result<Vec<HashSet<Value>>, Self::Error> {
+        let mut search_futures = Vec::new();
+        let n = keywords.len();
+
+        let buffered_memory = BatcherArc::new(MemoryBatcher::new_reader(
+            self.memory.clone(),
+            AtomicUsize::new(n),
+        )); // todo
```
tbrezot: todo

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+                        *self.decode,
+                    );
+
+                // Execute the search
+                findex.search(keyword).await
+            };
+
+            search_futures.push(future);
+        }
+        // at this point nothing is polled yet
```
tbrezot: `poll` is a implementation detail, please use the proper wording: `at this time no future is being processed` or something like that. But I actually do not think this comment is useful as you are simply explaining how Rust `async` works.
```suggestion
```

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+        // Execute all futures concurrently and collect results
+        let results = futures::future::join_all(search_futures).await;
+
+        // Process results
+        let mut output = Vec::with_capacity(results.len());
+        for result in results {
+            output.push(result?);
+        }
+
+        Ok(output)
```
tbrezot: just map and collect instead

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+    ) -> Result<(), Self::Error> {
+        self.batch_insert_or_delete(entries, true).await
+    }
+
+    async fn batch_delete(
+        &self,
+        entries: Vec<(Keyword, impl Sync + Send + IntoIterator<Item = Value>)>,
+    ) -> Result<(), Self::Error> {
+        self.batch_insert_or_delete(entries, false).await
+    }
```
tbrezot: Why making the difference between a search and an insert or a delete? The batching performs the same. Either uniformize the processing or inline `batch_insert_or_delete`.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+// The underlying tests assume the existence of a `Findex` implementation that is correct
+// The testing strategy for each function is to use the `Findex` implementation to perform the same operations
+// and compare the results with the `BatcherFindex` implementation.
+#[cfg(test)]
+mod tests {
```
tbrezot: Write a single memory-generic test instead that performs something non-trivial.

crate/findex/src/benches.rs
```diff
@@ -3,7 +3,10 @@
 //! to be generic and work with any memory back end that implements the MemoryADT
 //! trait.
 
-use crate::{Findex, IndexADT, MemoryEncryptionLayer, WORD_LENGTH, dummy_decode, dummy_encode};
+use crate::{
+    Findex, IndexADT, WORD_LENGTH, dummy_decode, dummy_encode,
+    encryption_layer::MemoryEncryptionLayer,
+};
```
tbrezot: why this change?

crate/findex/src/error.rs
```diff
@@ -16,3 +16,51 @@ impl<Address: Debug> Display for Error<Address> {
...
+    use super::*;
+    use cosmian_sse_memories::MemoryADT;
+
+    use crate::BatchingLayerError;
+
+    #[derive(Debug)]
+    pub enum BatchFindexError<M: MemoryADT> {
+        BatchingLayer(BatchingLayerError<M>),
+        Findex(Error<M::Address>),
+        Encoding(String),
```
tbrezot: Encoding errors should be caught by Findex and should not surface here.

crate/findex/src/error.rs
```diff
@@ -16,3 +16,51 @@ impl<Address: Debug> Display for Error<Address> {
...
+    }
+
+    impl<M: MemoryADT + Debug> std::error::Error for BatchFindexError<M> where
+        <M as MemoryADT>::Address: Debug
+    {
+    }
+}
+
+#[cfg(feature = "batch")]
+pub use batch_findex_error::*;
```
tbrezot: Move this at the top of the file to make more visible.

crate/findex/src/lib.rs
```diff
@@ -7,25 +7,33 @@
...
     Decoder, Encoder,
     generic_encoding::{generic_decode, generic_encode},
 };
-pub use encryption_layer::{KEY_LENGTH, MemoryEncryptionLayer};
 pub use error::Error;
-pub use findex::Findex;
-pub use findex::Op;
+pub use findex::{Findex, Op};
+
+pub use memory_layers::encryption_layer;
```
tbrezot: No! Do not change the visibility rules of the repo!

tbrezot (changes requested):
Okay, i did not go through with the reviewing: this MR is not in an acceptable state. Please, review your work and especially your diffs before submitting again, with a particular focus on:
- comments: do not repeat the code but if necessary explain the rationals behind it, spell check your comments, cleanup dead comments etc.
- naming: think through your naming convention to make it uniform and coherent. We talked about some names previously, please adopt them.
- do not modify files you do not need to modify! In particular be careful of the import statements.

I am asking a more-than-average code quality for the core Findex repository since it goes along with a published paper and its quality has an impact on our image. In the present state, this MR looks like a draft and is therefore not admissible for merging.

crate/findex/examples/databases.rs
HatemMn: `crate/findex/src/memory_layers` I added this **internal** module in order to "englobe" the batching layer and the encryption layer and whatever other layer that will be added, making the codebase more easely readable

crate/findex/examples/in_memory.rs
HatemMn: crate/findex/src/memory_layers I added this internal module in order to "englobe" the batching layer and the encryption layer and whatever other layer that will be added, making the codebase more easely readable

crate/findex/src/adt.rs
HatemMn: TODO

I had a reason actually :  it seemed more logical for me to have : 

The vector --> the index --> the batched index

if you don't agree comment yes and i'll revert it

crate/findex/src/adt.rs
HatemMn: I don't remember this name but I guess `IndexBatcher` isn't a bad idea, I will change it

crate/findex/src/adt.rs
HatemMn: True that's a wrong comment, thanks for the notice

I changed it to 
```
/// This trait  extends the functionality of the standard `IndexADT` by providing methods that operate
/// on multiple keywords or entries simultaneously to cut network's overhead and improve performance.
```

crate/findex/src/adt.rs
HatemMn: missed it, done

crate/findex/src/adt.rs
HatemMn: I just copy pasted the original index adt comment : 
image

crate/findex/src/batcher_findex.rs
HatemMn: fixes

crate/findex/src/batcher_findex.rs
HatemMn: I deleted that comment it's useless but I made this manual conversion for a reason : if findex (singular, non batch !) fails, I don't want to lose the error, and also, I don't want it to be "hidden" within a join error ( line 84, `futures::future::try_join_all(search_futures).await?;` ) - What I want is stopping everything and returning the original error

I might not be understanding the rust propagation correctly ? If you understood and agree with my comment close this otherwise I'm open to redicuss my approach

crate/findex/src/batcher_findex.rs
HatemMn: everything under `crate/findex/src/memory_layers/batching_layer`

To be more precise, "the batching layer" is that "fake" memory that takes a lot of BR/garded_write and transforms them to 1 single network call

crate/findex/src/batcher_findex.rs
HatemMn: renamed to "futures" to avoid useless naming hassle, thanks for the notice

crate/findex/src/batcher_findex.rs
HatemMn: TBH I made this version without regards this matter and plan to think about the runtime problems once we finish discussion the algorithms, as it will introduce even more complexities and retain us from examining the more technically challenging parts - I should have mentionned this in my pr message sorry !

But no worries, I intentionally made it this way and am aware of what you said

crate/findex/src/batcher_findex.rs
HatemMn: > Write a single memory-generic test instead that performs something non-trivial.

Those are very basic unit tests. If one function has issues (say, delete) we want a test that instantly says "delete is wrong/bugged". A general integration test should rather (as you said) perform non trivial things and notify us upon failing that we have a logic issue

This "non trivial test" is also from the things I am planning next, I will make a list in my PR description

crate/findex/src/benches.rs
HatemMn: `cargo fmt --all` formats this, I don't really have an opinion on the matter

crate/findex/src/error.rs
HatemMn: ok done

crate/findex/src/lib.rs
HatemMn: TODO

Are you sure that **this** was intended ?

```toml
pub use findex::Findex;
pub use findex::Op;
```

If yes can you explain me the motivation ? It seemed like a non intended mistake

HatemMn (commented):
Partial reply, marked the comments that need thinking with a 🚀 and treated the (mostly) trivial matters + added a "roadmap" to the pr description

crate/findex/README.md
HatemMn: New comment : 

> Findex also supports batching operations into a singe call to the memory interface, which reduces connection overhead and avoids file descriptor limits on some Linux systems.

crate/findex/src/batcher_findex.rs
HatemMn: done

crate/findex/src/batcher_findex.rs
HatemMn: fixed

crate/findex/src/batcher_findex.rs
HatemMn: ok

crate/findex/src/batcher_findex.rs
HatemMn: Used try_join_all , less lines, more effective

crate/findex/src/batcher_findex.rs
HatemMn: the operations are only similar on their first line - For the rest, I don't currently see how to refactor more

crate/findex/src/error.rs
HatemMn: after some digging, I notice that my error encoding type that I pass down is way too restrictive, I will work on giving him the bonds he actually need and only them

HatemMn: Done,

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,219 @@
...
+    // ```rust
+    // let mut aggregated_reads_results = self.inner.batch_read(all_addresses).await?;
+    //```
+    // This performs the *actual* call to the memory (example, a Redis call).
+    // This is different from the implementation of batch_read that is written below, which is just a fictive read
+    // operation that upper Findex(or whatever program) that uses this will be calling.
+    // In simple terms, we *need* two separate implementations of batch_read$
+    // This is way simpler to see for the write operation, as the guarded_write actual backend call will never
+    // happen in the first place. The manage function only calls batch_guarded_write.
+    pub inner: M,                     // The actual memory that does the R/W operations.
```
HatemMn: info for reviewers: there is no escape from the existence of the inner memory within this memory batcher.
     Proof: look at the manage function, this line in particular:
     ```rust
    let mut aggregated_reads_results = self.inner.batch_read(all_addresses).await?;
    ```
     This performs the *actual* call to the memory (example, a Redis call).
     This is different from the implementation of batch_read that is written below, which is just a fictive read
    operation that upper Findex(or whatever program) that uses this will be calling.
     In simple terms, we *need* two separate implementations of batch_read$
     This is way simpler to see for the write operation, as the guarded_write actual backend call will never
    happen in the first place. The manage function only calls batch_guarded_write.

Cargo.toml
```diff
@@ -30,3 +30,4 @@ cosmian_crypto_core = { version = "10.2", default-features = false, features = [
 criterion = { version = "0.6" }
 smol-macros = { version = "0.1" }
 tokio = { version = "1.46" }
+futures = "0.3.31" # Todo: this is a temporary approach
```
HatemMn: this will be taken away on the next PR (or at least i hope so)

crate/findex/src/findex.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(clippy::type_complexity)]
```
HatemMn, line 1: an outdated lint, for some reason

Cargo.toml
```diff
@@ -3,7 +3,7 @@ members = ["crate/findex", "crate/memories"]
 resolver = "2"
 
 [workspace.package]
-version = "8.0.1"
+version = "8.0.2"
```
tbrezot: This should be a minor version, not a debug one.

crate/findex/Cargo.toml
```diff
@@ -15,14 +15,15 @@ name = "cosmian_findex"
...
 
 [features]
+batch = ["cosmian_sse_memories/batch"]
 test-utils = ["agnostic-lite", "criterion"]
 
 [dependencies]
 aes = "0.8"
 cosmian_crypto_core.workspace = true
-cosmian_sse_memories = { path = "../memories", version = "8.0" }
+cosmian_sse_memories = { path = "../memories", version = "8.0.2" } # Should be above 8.0.2 to have batching features.
```
tbrezot: Using a minor will improve this.

crate/findex/README.md
```diff
@@ -2,6 +2,8 @@
 
 This crate provides the core functionality of Findex, defining the abstract data types, cryptographic operations, and encoding algorithms.
 
+Findex also supports batching operations into a singe call to the memory interface, which reduces connection overhead and avoids file descriptor limits on some Linux systems.
```
tbrezot: Batching shouldn't be Findex-specific.

crate/findex/src/adt.rs
```diff
@@ -51,6 +51,32 @@ pub trait VectorADT: Send {
     fn read(&self) -> impl Send + Future<Output = Result<Vec<Self::Value>, Self::Error>>;
 }
 
+/// This trait  extends the functionality of the standard `IndexADT` by
+/// providing methods that operate on multiple keywords or entries
+/// simultaneously.
+#[cfg(feature = "batch")]
+pub trait IndexBatcher<Keyword, Value> {
```
tbrezot: You say this is an extension, but this trait is not a super-trait of `IndexADT`. Maybe use another wording.

crate/findex/src/adt.rs
```diff
@@ -51,6 +51,32 @@ pub trait VectorADT: Send {
...
+    type Error: std::error::Error;
+
+    /// Search the index for the values bound to the given keywords.
+    fn batch_search(
+        &self,
+        keywords: Vec<&Keyword>,
```
tbrezot, line 63: Best not to use a vector to avoid imposing needless collection upon the client.

crate/findex/src/adt.rs
```diff
@@ -51,6 +51,32 @@ pub trait VectorADT: Send {
...
+    ) -> impl Future<Output = Result<Vec<HashSet<Value>>, Self::Error>>;
```
tbrezot, line 64: Hasn't "async in trait" become stable by now? Maybe try with `async fn batch_search(...) -> Result<..., ...>;`

crate/findex/src/adt.rs
```diff
@@ -51,6 +51,32 @@ pub trait VectorADT: Send {
...
+
+    /// Binds each value to their associated keyword in this index.
+    fn batch_insert(
+        &self,
+        entries: Vec<(Keyword, impl Sync + Send + IntoIterator<Item = Value>)>,
```
tbrezot: Are both `Sync` and `Send` really needed here?

crate/findex/src/findex_batcher.rs
```diff
@@ -0,0 +1,299 @@
...
+use crate::{Decoder, Encoder, Findex, IndexADT, adt::IndexBatcher, error::BatchFindexError};
+
+#[derive(Debug)]
+pub struct FindexBatcher<
+    const WORD_LENGTH: usize,
+    Value: Send + Hash + Eq,
+    EncodingError: Send + Debug,
+    BatcherMemory: Clone + Send + BatchingMemoryADT<Address = Address<ADDRESS_LENGTH>, Word = [u8; WORD_LENGTH]>,
+> {
+    memory: BatcherMemory,
```
tbrezot, line 14:
```suggestion
    memory: MemoryBatcher,
```

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,299 @@
...
+    ) -> Self {
+        Self {
+            memory,
+            encode,
+            decode,
+        }
+    }
+
+    // Both insert and delete operations make an unbounded number of calls to
+    // `guarded_write` on the memory layer.
```
tbrezot: I am not sure why you need this comment here. The Findex paper analyses the performance of Findex operations and your comment just rewords a consequence of the lock-free property.

crate/findex/src/findex_batcher.rs
```diff
@@ -0,0 +1,299 @@
...
+            decode,
+        }
+    }
+
+    // Both insert and delete operations make an unbounded number of calls to
+    // `guarded_write` on the memory layer.
+    async fn batch_insert_or_delete<Keyword>(
+        &self,
+        entries: Vec<(Keyword, impl Send + IntoIterator<Item = Value>)>,
+        is_insert: bool,
```
tbrezot, line 45: Can you pass a function instead? This would allow reusing this function for searches too.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,299 @@
...
+    async fn batch_insert_or_delete<Keyword>(
+        &self,
+        entries: Vec<(Keyword, impl Send + IntoIterator<Item = Value>)>,
+        is_insert: bool,
+    ) -> Result<(), BatchFindexError<BatcherMemory>>
+    where
+        Keyword: Send + Sync + Hash + Eq,
+    {
+        let mut futures = Vec::new();
```
tbrezot: Allocate the memory up-front using `with_capacity()`.

crate/findex/src/findex_batcher.rs
```diff
@@ -0,0 +1,299 @@
...
+        let memory = Arc::new(MemoryBatcher::new(self.memory.clone(), entries.len()));
+
+        for (guard_keyword, bindings) in entries {
+            let memory = memory.clone();
+            // Create a temporary Findex instance using the shared batching layer.
+            let findex = Findex::<WORD_LENGTH, Value, EncodingError, _>::new(
+                memory.clone(),
+                self.encode,
+                self.decode,
+            );
```
tbrezot, line 69: Why is this created once per iteration? Maybe create it once and for all outside of the loop instead.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,299 @@
...
+                self.decode,
+            );
+
+            let future = async move {
+                if is_insert {
+                    findex.insert(guard_keyword, bindings).await
+                } else {
+                    findex.delete(guard_keyword, bindings).await
+                }?;
+                // Once one of the operations succeeds, we should make the buffer smaller.
```
tbrezot: The concept of buffer is an internal of the memory batcher, which makes this comment hard to understand without knowing the memory batcher implementation.
```suggestion
                // Once an operation succeeds, the batch size needs to be reduced.
```

crate/findex/src/findex_batcher.rs
```diff
@@ -0,0 +1,299 @@
...
+
+            let future = async move {
+                if is_insert {
+                    findex.insert(guard_keyword, bindings).await
+                } else {
+                    findex.delete(guard_keyword, bindings).await
+                }?;
+                // Once one of the operations succeeds, we should make the buffer smaller.
+                memory.unsubscribe().await?;
+                Ok::<_, BatchFindexError<_>>(())
```
tbrezot, line 81: Isn't this what `unsusribe()` returns? Maybe this line is not needed.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,299 @@
...
+    async fn batch_insert_or_delete<Keyword>(
+        &self,
+        entries: Vec<(Keyword, impl Send + IntoIterator<Item = Value>)>,
+        is_insert: bool,
+    ) -> Result<(), BatchFindexError<BatcherMemory>>
+    where
+        Keyword: Send + Sync + Hash + Eq,
+    {
+        let mut futures = Vec::new();
+        let memory = Arc::new(MemoryBatcher::new(self.memory.clone(), entries.len()));
```
tbrezot: It may be clearer if the memory batcher API requires us to make an explicit call to some subscription function to have more parallelism with the call to `unsubscribe`: `memory.subscribe(entries.len())`. Doing so may have drawbacks and is therefore open for discussion.

crate/findex/src/findex_batcher.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        for keyword in keywords {
+            let memory = memory.clone();
+            let findex = Findex::<WORD_LENGTH, Value, EncodingError, _>::new(
+                memory,
+                self.encode,
+                self.decode,
+            );
+
+            let future = async move { findex.search(keyword).await };
```
tbrezot, line 151: This code is only correct under the assumption that the number of memory calls performed by Findex search is the same across all the calls. This is true, but needs to be documented for clarity, otherwise one can wonder why calling `unsubscribe()` is not required here while it is for `insert` and `delete`.

crate/findex/src/batcher_findex.rs
```diff
@@ -0,0 +1,299 @@
...
+        futures::future::try_join_all(futures)
+            .await
+            .map_err(|e| BatchFindexError::Findex(e))
+    }
+}
+
+// The underlying tests assume the existence of a `Findex` implementation that
+// is correct The testing strategy for each function is to use the `Findex`
+// implementation to perform the same operations and compare the results with
+// the `BatcherFindex` implementation.
```
tbrezot:
```suggestion
// These tests implement dual testing against the base Findex implementation.
```

crate/findex/src/lib.rs
```diff
@@ -30,6 +30,13 @@ pub use encryption_layer::{KEY_LENGTH, MemoryEncryptionLayer};
 pub use error::Error;
 pub use findex::{Findex, Op};
 
+#[cfg(feature = "batch")]
+mod batcher_findex;
+#[cfg(feature = "batch")]
+pub use adt::IndexBatcher;
+#[cfg(feature = "batch")]
+pub use batcher_findex::FindexBatcher;
```
tbrezot:
```suggestion
pub use findex_batcher::FindexBatcher;
```

crate/memories/src/batching_layer/buffer.rs
```diff
@@ -0,0 +1,95 @@
+use std::{mem, sync::Mutex};
+
+use crate::{
+    BatchingMemoryADT,
+    batching_layer::operation::{Operation, PendingOperations},
+};
+
+struct Buffer<M: BatchingMemoryADT> {
+    capacity: usize, // the size at which the buffer should be flushed
```
tbrezot: I think the term "capacity" is clear enough. What is lacking, however, is a module-level doc explaining the inner-workings of the memory batcher.

```suggestion
    capacity: usize,
```

crate/memories/src/batching_layer/buffer.rs
```diff
@@ -0,0 +1,95 @@
...
+        Ok(buffer.flush_if_not_empty())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BufferError {
+    TypeMismatch, // when the type of the new item does not match the type of the last item
+    Overflow,
+    Underflow,
+    Mutex(String),
```
tbrezot: Simply unwrap on mutex error.

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+
+impl<M: BatchingMemoryADT + Send + Debug> MemoryBatcher<M>
+where
+    M::Address: Clone + Send,
+    M::Word: Send + std::fmt::Debug,
+{
+    pub fn new(inner: M, n: usize) -> Self {
+        if n == 0 {
+            panic!("Buffer capacity must be greater than zero.");
+        };
```
tbrezot: You can use a `NonZeroUsize` instead. Also, name this parameter `capacity`.

crate/memories/src/batching_layer/buffer.rs
```diff
@@ -0,0 +1,95 @@
...
+        // Check if the new item is compatible with the last item, since the buffer is
+        // thread-safe, this ensures by transitivity that all items in the
+        // buffer are of the same type.
+        if let Some(last_item) = buffer.data.last() {
+            if mem::discriminant(last_item) != mem::discriminant(&item) {
+                return Err(BufferError::TypeMismatch);
+            }
+        }
+        buffer.data.push(item);
+        Ok(buffer.flush_if_not_empty())
```
tbrezot, line 64: It seems that all this function does is to return its argument : since you are still holding the lock, calls to push are sequential and all finish with emptying the buffer.

crate/memories/src/batching_layer/buffer.rs
```diff
@@ -0,0 +1,95 @@
...
+    Mutex(String),
+}
+
+impl<M: BatchingMemoryADT> From<std::sync::PoisonError<std::sync::MutexGuard<'_, Buffer<M>>>>
+    for BufferError
+{
+    fn from(e: std::sync::PoisonError<std::sync::MutexGuard<'_, Buffer<M>>>) -> Self {
+        Self::Mutex(format!("Mutex poisoned: {}", e))
+    }
+}
```
tbrezot: Simply unwrap upon lock poisoning:
```suggestion
}
```

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    async fn guarded_write(
+        &self,
+        guard: (Self::Address, Option<Self::Word>),
+        bindings: Vec<(Self::Address, Self::Word)>,
+    ) -> Result<Option<Self::Word>, Self::Error> {
+        (**self).guarded_write(guard, bindings).await
+    }
+}
```
tbrezot: Can't you make your memory batcher sharable instead ?

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+                )
+            }
+            MemoryInput::Write((guard, bindings)) => {
+                let (sender, receiver) = oneshot::channel();
+                (
+                    Operation::Write(((guard, bindings), sender)),
+                    OperationResultReceiver::<M>::Write(receiver),
+                )
+            }
+        };
```
tbrezot, line 102: I think you can factor out the channel creation.

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+        Ok(match receiver {
+            OperationResultReceiver::Read(receiver) => {
+                let result = receiver.await?.map_err(MemoryBatcherError::Memory)?;
+                MemoryOutput::Read(result)
+            }
+            OperationResultReceiver::Write(receiver) => {
+                let result = receiver.await?.map_err(MemoryBatcherError::Memory)?;
+                MemoryOutput::Write(result)
+            }
+        })
```
tbrezot, line 117: Same, you should be able to factor out the result reception.

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    async fn manage(&self, ops: PendingOperations<M>) -> Result<(), MemoryBatcherError<M>> {
+        // Assumes the vector is homogeneous, i.e. all operations are of the same type.
+        // This should be guaranteed by the buffer.
+        match ops[0] {
+            Operation::Read(_) => {
```
tbrezot, line 124: You may want to avoid querying the same address once.

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+                // Build combined address list while tracking which addresses belong to which
+                // batch.
+                let all_addresses: Vec<_> = ops
+                    .iter()
+                    .flat_map(|op| match op {
+                        Operation::Read((addresses, _)) => addresses.clone(),
+                        _ => unreachable!(
+                            "Expected all operations to be reads, reaching this statement means \
+                             the buffer has implementation flaws at the push level."
+                        ),
```
tbrezot: Maybe return an error instead. If you do so, you should remove the homogeneity check in `push`.

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+                        Operation::Read((addresses, sender)) => (addresses, sender),
+                        _ => unreachable!(
+                            "Expected all operations to be reads, reaching this statement means \
+                             the buffer has implementation flaws at the push level."
+                        ),
+                    })
+                    .rev()
+                {
+                    let split_point = aggregated_reads_results.len() - input_addresses.len(); // This is the point where the last batch's results start.
+                    let batch_results = aggregated_reads_results.split_off(split_point); // After this call, all_results will be left containing the elements [0, split_point).
```
tbrezot: Shorter names allow for clearer code:
```suggestion
                    let batch_words = words.split_off(words.len() - batch_addresses.len());
```

crate/memories/src/batching_layer/memory.rs
```diff
@@ -0,0 +1,217 @@
...
+                    sender.send(Ok(batch_results)).map_err(|_| {
+                        // Upon failure, the vector we tried to send is returned in the Err variant,
+                        // but it's explicitly ignored here to not extract information.
+                        MemoryBatcherError::<M>::Channel(
+                            "The receiver end of this read operation was dropped before the \
+                             `send` function could be called."
+                                .to_owned(),
+                        )
```
tbrezot:
```suggestion
                        MemoryBatcherError::<M>::ClosedChannel
```

crate/memories/src/batching_layer/operation.rs
```diff
@@ -0,0 +1,62 @@
...
+
+pub(crate) type ReadOperation<M> = (
+    BatchReadInput<M>,
+    oneshot::Sender<Result<BatchReadOutput<M>, <M as MemoryADT>::Error>>,
+);
+
+pub(crate) type WriteOperation<M> = (
+    GuardedWriteInput<M>,
+    oneshot::Sender<Result<GuardedWriteOutput<M>, <M as MemoryADT>::Error>>,
+);
```
tbrezot, line 48: Is it a write or a guarded write?

crate/memories/src/lib.rs
```diff
@@ -54,3 +54,22 @@ pub trait MemoryADT {
...
+#[cfg(feature = "batch")]
+pub trait BatchingMemoryADT: MemoryADT {
+    #[allow(clippy::type_complexity)] // Refactoring this type will make the code unnecessarily more difficult to read without any actual benefit.
+    fn batch_guarded_write(
+        &self,
+        write_operations: Vec<(
+            (Self::Address, Option<Self::Word>),
+            Vec<(Self::Address, Self::Word)>,
+        )>,
+    ) -> impl Send + std::future::Future<Output = Result<Vec<Option<Self::Word>>, Self::Error>>;
```
tbrezot: You define the appropriate type in `operation.rs`:
```suggestion
    fn batch_guarded_write(
        &self,
        write_operations: Vec<GuardedWriteInput<Self>>,
    ) -> impl Send + std::future::Future<Output = Result<Vec<GuardedWriteOutput<Self>>, Self::Error>>;
```

crate/memories/src/in_memory.rs
```diff
@@ -74,6 +76,30 @@ impl<Address: Send + Hash + Eq + Debug, Value: Send + Clone + Eq + Debug> Memory
...
+            let (a, old) = guard;
+            let cur = store.get(&a).cloned();
+            if old == cur {
+                for (k, v) in bindings {
+                    store.insert(k, v);
+                }
+            }
+            res.push(cur);
+        }
+        Ok(res)
```
tbrezot: Use a `map` instead.

tbrezot (commented):
There still may have some issues with the core logic. See my comments about the push function.

Otherwise, this is mostly style and readability improvements.

Cargo.toml
HatemMn: fixed

crate/findex/Cargo.toml
HatemMn: idk why but I don't have this comment on my files

I deleted it anyway

crate/findex/README.md
HatemMn: I tried to change the sentence but can't find a better equivalent, do you suggest one ? Or I should delete this line ?

In my personnal opinion nothing in this sentence insinuates a tied-relashionship between findex and the batching layer

crate/findex/src/adt.rs
HatemMn: new comment : 
/// This trait provides methods that let an index operate on multiple keywords or entries simultaneously.

crate/findex/src/adt.rs
HatemMn: image

I think I tried that, and probably stopped because of this. Not because we can't use lifetime specefier, but more because I think we cummonly try to avoid their usage

HatemMn: Thanks to github bugs, A comment I wrote here was lost

I remember saying that we had 3 solutions : 
- keep it as is
- accept to add lifetimes to the definition
- take ownership of the Keywords (I do not prefer doing this as it will induce non necessary cloning )

close this if you think we should keep it like it now

### By the way,I agree with your comment and that I have implemented it for the other functions (batch_insert, etc...)

crate/findex/src/adt.rs
HatemMn: Hell yeah looks like it, I will update the adequate traits https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/

UPDATE : this won't be possible with all traits on this code since the async in trait does not garantee the send - I did a try on the branch called **test/async_trait** , which is pushed to the remot

crate/findex/src/adt.rs
HatemMn: Just a general note

I found myself also obliged to add "ExactSizeIterator" bound - basically garantees that we can call **len()** . This is mandatory, to know how big the memory should be instantiated

```rust
    fn batch_insert<Values, Entries>(
        &self,
        entries: Entries,
    ) -> impl Send + Future<Output = Result<(), Self::Error>>
    where
        Values: Sync + Send + IntoIterator<Item = Value>,
        Entries: Send + IntoIterator<Item = (Keyword, Values)>,
        Entries::IntoIter: ExactSizeIterator;
```

crate/findex/src/adt.rs
HatemMn: I did a global check, only keywords need to be Sync, so yes we don't need both - I updated accordingly

crate/findex/src/findex_batcher.rs
HatemMn: There is a confusion that I fell in despite being the author of this problem
- **MemoryBatcher** : that "fictive" memory structure that conatins the buffer
- **BatcherMemory** : a generic type for index batcher implentations : 
```rust
impl<
    const WORD_LENGTH: usize,
    Keyword: Send + Sync + Hash + Eq,
    Value: Send + Hash + Eq,
    EncodingError: Send + Debug,
    BatcherMemory: Debug
        + Send
        + Sync
        + Clone
        + BatchingMemoryADT<Address = Address<ADDRESS_LENGTH>, Word = [u8; WORD_LENGTH]>,
> IndexBatcher<Keyword, Value> for FindexBatcher<WORD_LENGTH, Value, EncodingError, BatcherMemory>
{
```

BatcherMemory is basically "A memory that functions as a memoryBatcher", hence the name

I 100% Agree that this is confusing

I changed the name to **BatchingMemoryLayer** to avoid confusiong ( any meaningful names are ugly - I didn't find a lot of ideas )

crate/findex/src/batcher_findex.rs
HatemMn: It was written in order to explain why I refactored the insert & delete. I think the comment does not explain my intention so I will place it elsewhere or delete it

crate/findex/src/findex_batcher.rs
HatemMn: Sorry, I really tried to think but I do not understand this comment

Searches operate differently, If you don't mind having a middle big "if/else" block I can refactor the search here yeah, but I think we try to avoid those. If the goal is stricly less lines of code without regards to readness, then yes we can move the search here

crate/findex/src/batcher_findex.rs
HatemMn: done

crate/findex/src/findex_batcher.rs
HatemMn: I'm not sure but _I think_ we have already discussed this, the problem is the compiler having issues to infer the **error** return type

I don't like this line either but removing it triggers an error here that I need to address making the code even uglier

```rust
               if is_insert {
                    findex.insert(guard_keyword, bindings).await
                } else {
                    findex.delete(guard_keyword, bindings).await
                }?; // <------ This line
```

crate/findex/src/findex_batcher.rs
HatemMn: This is why I had that comment about _unbounded number of calls_ that we discussed in some comment above. Anyway I do agree with you that this isn't trivial to understand, i will update the comments accoprdingly

crate/findex/src/findex_batcher.rs
HatemMn: it's true that one **Arc<Findex>** that's cloned accross the loop does the job

however after reconsidereation I remembered why I used multiple Findexes and correct me if I'm wrong : 

If we have multiple instances of Findex, if the number of available worker threads permits it, they will be able to work in parallel. However, if a single findex if shared, on each (internal) **guarded_write** call, the caller will not be able to work if a GW is already ongoing elsewhere

This also aligns with the initial drawing we did when we drafted the algorithme (I will find it if you want it's stored locally on my photo backups)

I might be mistaken, since I do not have profound knowledge about the low lever functionning of async rust, so idk

 and since our tests are already fast it will be a bit hard to "test" my claims in an empirical manner... So please tell me if you agree about my idea

crate/findex/src/batcher_findex.rs
HatemMn: done

crate/findex/src/batcher_findex.rs
HatemMn: done

crate/memories/src/batching_layer/buffer.rs
HatemMn: okay DONE... but why ?

we will probably face some clippy consequences (that I will silence, for the matter)

crate/memories/src/batching_layer/memory.rs
HatemMn: sadly, it's not possible, I remember trying and finally accepting that those two channels have different types (that hopefully for us the compiler is smart enough to infer)

so it needs to not be refactored (if you try to do it the program won't compile)

crate/memories/src/batching_layer/memory.rs
HatemMn: same as above, this isn't possible (I tried and to the extent of what I know I could not make it compile)

crate/memories/src/batching_layer/memory.rs
HatemMn: I see what you mean but isn't this supposed to be implemented on the batch_read level ? Or even left to the user's choice to do it or not ? If you think still that we should do this, I think this whole algorithm design should be changed, I propose the following : 
- Delete all this read algorithm
- iterate over **ops: PendingOperations<M>** (who is basically a vector of those : 
```rust
pub(crate) type ReadOperation<M> = (
    BatchReadInput<M>,
    oneshot::Sender<Result<BatchReadOutput<M>, <M as MemoryADT>::Error>>,
);
```
Let's call `i` the iteration number and 'j' the nested address iterator
- Create and populate two hashMaps : 
    - First hashmap will be address[j] --> i
    - Second hashmap will be i --> sender[i]  ( sender is the channel )
- Now, we collect all the addresses in some Set structure (HashSet, for instance)
- Perform the BR : 
```rust
             let mut aggregated_reads_results =           self.inner
                    .batch_read(all_addresses)
                    .await
                    .map_err(MemoryBatcherError::Memory)?;
```
- Finally, we iterate over the results and assign them back to their sender by using HasMap2[ HashMap1[ adr ] ]
- Once the result verctor reconstructed we send the results

Now, this method is slower time wise and more space hungry space wise, findex's bottlneck is the network calls and not the address repetition, do we really want such an algorithm ?

crate/memories/src/batching_layer/memory.rs
HatemMn: The new code looks more complex but it's necessary, the `unreachable!` seemed more clear and since it was (really) unreachable we could avoid the double collect (that's basically like a double iteration, right ?). Putting an error looks like it "can" happen

Here is the new committed code, please tell me what to keep
```rust
let all_addresses: Vec<_> = ops
                    .iter()
                    .map(|op| match op {
                        Operation::Read((addresses, _)) => Ok(addresses.clone()),
                        _ => Err(MemoryBatcherError::Buffering(
                            crate::batching_layer::buffer::BufferError::TypeMismatch,
                        )),
                    })
                    .collect::<Result<Vec<_>, _>>()? // This will short-circuit on first error
                    .into_iter()
                    .flatten()
                    .collect();
```

The previous code : 

```rust
let all_addresses: Vec<_> = ops
                    .iter()
                    .flat_map(|op| match op {
                        Operation::Read((addresses, _)) => addresses.clone(),
                        _ => unreachable!(
                            "Expected all operations to be reads, reaching this statement means \
                             the buffer has implementation flaws at the push level."
                        ),
                    })
                    .collect();
```

HatemMn: #### IMHO I vouch for the previous method, costing an O(1) and garanteeing no weird behaviors vs. the current one that's 3 times O(N) (collect), makes the code longer, and permits a scenario with heterogenous vector, but I will let you chose

HatemMn: to avoid comments spam, the two messages above also apply for this too : 
```rust
 // Distribute results to each batch's sender.
                for (input_addresses, sender) in ops
                    .into_iter()
                    .map(|op| match op {
                        Operation::Read((addresses, sender)) => (addresses, sender),
                        _ => unreachable!(
                            "Expected all operations to be reads, reaching this statement means \
                             the buffer has implementation flaws at the push level."
                        ),
                    })
                    .rev()
```

and also this one : 

```rust
let (bindings, senders): (Vec<_>, Vec<_>) = ops
                    .into_iter()
                    .map(|op| match op {
                        Operation::Write((bindings, sender)) => (bindings, sender),
                        _ => unreachable!(),
                    })
                    .unzip();
```

same approach and same considerations

crate/memories/src/batching_layer/memory.rs
HatemMn: Okay for **words** but I honestly had trouble understanding what `batch_addresses` mean so I kept "input"
I also had to delete the comment above as you suggested

crate/memories/src/batching_layer/memory.rs
HatemMn: done

crate/memories/src/batching_layer/operation.rs
HatemMn: It's a "write", and to be exact it's a "Bulk of guarded writes" bundled together, which can also be called "batch_guarded_write" as the trait name

This looked too verbose tho, so I just said "write"

This isn't a public type anyway

crate/memories/src/lib.rs
HatemMn: That type was not public - I will make it public then

crate/memories/src/in_memory.rs
HatemMn: done

crate/memories/src/batching_layer/memory.rs
HatemMn: I deleted this and made it shareable by implementing the "clone" myself and putting the two inner variables within Arc<...>

crate/memories/src/batching_layer/buffer.rs
HatemMn: done

crate/findex/README.md
HatemMn: i deleted the word findex

crate/memories/src/batching_layer/memory.rs
HatemMn: done

HatemMn (commented):
Some changes were made according to the comments. Some of the "huge" changes were said in replies but not implemented till getting approval for them

This set of commits covers strictly the comments of the previous review, further commits might follow for the rest of the tasks (notably, getting rid of the `futures` crate that was there mostly for POC purposes)

Also please check out my alternative for the `unreachable!` statement and tell me if I shall revert the previous version or not, as the new one seems ( IMO )  worse of what was there

PR #155, opened 2025-08-14
[Style] Cargo fmt update

HatemMn:
Very minor PR

I noticed that cargo fmt file has not been changed since 3 years and is out of date with other repositories

In this PR I only changed (manually) the file **.rustfmt.toml** + updated the cargo toml version + the ci on this pr : https://github.com/Cosmian/reusable_workflows/pull/54

By the way, the `version` option is deprecated. I use `style_edition` instead.

All the other diffs are stricly auto reformatted

## ⚠️  Notice

Oddly enough some rules were written in that file yet not applied to the repo.

Example : 
pub use findex::Findex;
pub use findex::Op;

should have been merged in 1 line.

After reading the CI, exactly here : 

image

Do notice that the **rust fmt** check uses the default available channel to check formatting **but almost of the rules we use are from the nightly channel**

In order to avoid falsly green CIs, I will proceed to make an update to use the nightly channel ONLY for formatting via 
```bash
cargo +nightly fmt --all -- --check --color always
 ```

IMO as long as the CI continues to mistake the formatting channel, we will always have this problem
image

Workflow link : https://github.com/Cosmian/findex/actions/runs/16972645804/job/48113380578?pr=155

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on: push
 
 jobs:
   cargo-lint:
-    uses: Cosmian/reusable_workflows/.github/workflows/cargo-nursery.yml@develop
+    uses: Cosmian/reusable_workflows/.github/workflows/cargo-nursery.yml@feat/format_nightly_only
     with:
       toolchain: 1.87.0
       workspace: true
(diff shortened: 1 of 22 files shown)


HatemMn:
suceeded by : https://github.com/Cosmian/findex/pull/156

PR #156, opened 2025-08-18
ci: verify cargo fmt in nightly

Manuthor:


.github/workflows/ci.yml
```diff
@@ -5,14 +5,43 @@ on: push
...
+  cargo-doc:
+    uses: Cosmian/reusable_workflows/.github/workflows/cargo-doc.yml@develop
+    with:
+      toolchain: 1.87.0
+  cargo-semver:
+    uses: Cosmian/reusable_workflows/.github/workflows/cargo-semver.yml@develop
+    with:
+      toolchain: 1.87.0
+  cargo-publish-workspace:
+    uses: Cosmian/reusable_workflows/.github/workflows/cargo-publish-workspace.yml@develop
```
HatemMn: is there a sort of way to refactor that toolchain argument ? It's much likely to be changed and rechanged

Not quite sure of the syntax but I think smth of the sort is possible ?  :
 
```
env:
  RUST_TOOLCHAIN: "1.87.0"
  
jobs:
  cargo-lint:
    uses: Cosmian/reusable_workflows/.github/workflows/cargo-nursery.yml@develop
    with:
      toolchain: ${{ env.RUST_TOOLCHAIN }} # <------
      workspace: true
    secrets:
      token: ${{ secrets.CRATES_IO }}
  ```

HatemMn (commented):
Just a detail the version should be bumped to 8.0.1 in the cargo toml like here https://github.com/Cosmian/findex/pull/155/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542 , no ?

Otherwise I'm good with this, we have more granular control over this CI now

.github/workflows/ci.yml
Manuthor: unfortunately, it does not work

Manuthor: Take a look at 317e3aeccde818df014ac3fd23c9f364154fced5

p0wline (approved):

